PR #7, opened 2022-05-12
Add preliminary prototype of s2s TimeIndex

Peter9192:
Add a first prototype of what a dedicated S2S TimeIndex could look like.


geek-yang (approved):
Good job @Peter9192 ! Thanks for fixing all the tests. Regarding the descriptions in the top post of this PR, I think it is better to find a place to store it and keep tracking all of them in later iterations. I made an issue for it (#9) and we can address it in another PR. Thanks again for the nice work! Feel free to merge this one.

PR #10, opened 2022-05-18
Docs for developer

### LINKED ISSUE #9: Development/API design docs
We need to find a place to keep tracking the development description in all PRs when necessary (e.g. PR #7  ). A good option is to create a `docs` folder in the repo and store all development/API design and algorithm related docs there.

These docs can be used to generate a formal documentation using sphinx later.

geek-yang:
This PR closes #9.

We move all docs (except for README.md) to the docs folder and create a developer doc to keep tracking our thoughts and choices regarding the software design.

The main reason to move `.md` files from home to docs, is to have a cleaner repo home.


geek-yang:
> Looking at this PR, I started wondering whether it's actually considered good practice to have these .md files in the home of the repo, or whether it's also fine to move them to the docs. I couldn't really find any good resources on that.
> 
> I did find [dianna-ai/dianna#161](https://github.com/dianna-ai/dianna/issues/161) which is probably what inspired you to do the same here, right? From that discussion the reasoning and motivation is clear and I understand why you would want to do this. Perhaps it would be good to also include the same reasoning in this issue/PR? So: the main reason to move it from home to docs, is to have a cleaner repo home.

Thanks for the comment. That's right and I thought it would be much cleaner to put all the `.md` files in one place, which makes it both easier for the user to find the information in one location and for the developer to keep tracking and maintaining these docs.

I agree that it is good to include the reasoning here. I will put it in the top post.

docs/README.dev.md
```diff
@@ -1,6 +1,6 @@
 # `s2s` developer documentation
 
-If you're looking for user documentation, go [here](README.md).
+If you're looking for user documentation, go [here](../README.md).
```
BSchilperoort, line 3: This link seems to be broken in the sphinx generated documentation ([quick link](https://ai4s2s--10.org.readthedocs.build/en/10/README.dev.html)).

docs/project_setup.md
```diff
@@ -45,12 +45,12 @@ help you decide which tool to use for packaging.
   - PyTest is listed as a development dependency
   - This is configured in `setup.cfg`
 - The project uses GitHub action workflows to automatically run tests on GitHub infrastructure against multiple Python versions
-  - Workflows can be found in [`.github/workflows`](.github/workflows/)
+  - Workflows can be found in [`.github/workflows`](../.github/workflows/)
```
BSchilperoort, line 48: This reference is also broken in the generated docs

docs/project_setup.md
```diff
@@ -45,12 +45,12 @@ help you decide which tool to use for packaging.
...
 - [Relevant section in the guide](https://guide.esciencecenter.nl/#/best_practices/language_guides/python?id=testing)
 
 ## Documentation
 
-- Documentation should be put in the [`docs/`](docs/) directory. The contents have been generated using `sphinx-quickstart` (Sphinx version 1.6.5).
+- Documentation should be put in the [`docs/`](../docs/) directory. The contents have been generated using `sphinx-quickstart` (Sphinx version 1.6.5).
```
BSchilperoort, line 53: This reference is also broken in the generated docs

BSchilperoort (commented):
Hi Yang, the drop-down menus seem to be working fine, but I found a couple of references that seem to be broken in the generated docs.
The references work fine in the markdown files, but not on the html pages.

BSchilperoort (approved):
The documentation renders correctly now (as far as I can see). There is one warning remaining, about the autoapi pages missing from the toctree, but it is accessible and works fine.

geek-yang:
Thanks for the review and the changes made to the document. It looks awesome! I will merge this PR.

PR #18, opened 2022-06-05
Implement map year to data function

geek-yang:
As a user, I want to generate a calendar with preferred intervals and the time range should cover my data.


s2s/time.py
```diff
@@ -170,6 +174,35 @@ def map_years(
 
         return index
 
+    def map_year_to_data(self, input_data: Union[pd.Series, pd.DataFrame, xr.Dataset, xr.DataArray],
+        flat: bool = False
+        ) -> pd.DataFrame:
+        """Map the calendar to input data period.
+        
+        Get time range from input data and generate correpodning interval index 
+        using ``map_year`` and ``map_years`` functions.
```
Peter9192, line 183:
```suggestion
        Get time range from input data and generate corresponding interval index.
```

s2s/time.py
```diff
@@ -170,6 +174,35 @@ def map_years(
...
+
+        Args:
+            input_data: Input data for year mapping. Its index must be pandas.DatetimeIndex.
```
Peter9192, line 186: year-mapping - are we only mapping the years, or are we also accounting for the exact start/end date of the data.

tests/test_time.py
```diff
@@ -58,6 +58,51 @@ def test_map_years(self):
...
+        # single year in input data
+        time_index = pd.date_range('20210401', '20211231', freq='180d')
+        test_data = np.random.random(len(time_index))
+        timeseries = pd.Series(test_data, index=time_index)
+        year = cal.map_year_to_data(timeseries)
+
+        expected = np.array(
+            [
+                interval("2021-07-04", "2021-12-31"),
+                interval("2021-01-05", "2021-07-04"),
```
Peter9192, line 72: Note that this last interval is half empty, as your test data starts only in april. I think the package should warn the user about this.

tests/test_time.py
```diff
@@ -58,6 +58,51 @@ def test_map_years(self):
...
+            ]
+        )
+
+        assert np.array_equal(years, expected)
+
+        # timeindex reverse order
+        timeseries = pd.Series(test_data, index=time_index[::-1])
+        years = cal.map_year_to_data(timeseries)
+
+        assert np.array_equal(years, expected)
```
Peter9192, line 104: Add a test case for when the test data ends before the anchor date. E.g. anchor date is 30 Nov, test data ends in August 2020. Then the entire year 2020 should be discarded, but the current implementation will keep it.

s2s/time.py
```diff
@@ -170,6 +174,35 @@ def map_years(
 
         return index
 
+    def map_year_to_data(self, input_data: Union[pd.Series, pd.DataFrame, xr.Dataset, xr.DataArray],
```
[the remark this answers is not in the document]
geek-yang: Corrected!

s2s/time.py
```diff
@@ -170,6 +174,35 @@ def map_years(
...
+        Get time range from input data and generate correpodning interval index 
+        using ``map_year`` and ``map_years`` functions.
+
+        Args:
+            input_data: Input data for year mapping. Its index must be pandas.DatetimeIndex.
+            flat: Same as the argument in ``map_years``.
+
+        Returns:
+            Pandas DataFrame filled with Intervals of the calendar's frequency.
+        """
```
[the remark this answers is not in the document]
geek-yang: Good point! A small note is now added there.

s2s/time.py
geek-yang: Thanks for the suggestion. I rephrase it as "input_data: Input data for datetime mapping...", sounds much proper now.

s2s/time.py
geek-yang: Corrected!

tests/test_time.py
geek-yang: Thanks for the comment! Now tests concerning edge values are added.

tests/test_time.py
geek-yang: Thanks for the inspiration! A warning is added when parts of the generated intervals couldn't be covered by the input data.

s2s/time.py
```diff
@@ -170,6 +174,53 @@ def map_years(
 
         return index
 
+    def map_to_data(self, input_data: Union[pd.Series, pd.DataFrame, xr.Dataset, xr.DataArray],
+        flat: bool = False
+        ) -> pd.DataFrame:
+        """Map the calendar to input data period.
+        
+        Get datetime range from input data and generate correpodning interval index. This method
```
BSchilperoort, line 182: small typo here: `corresponding`

s2s/time.py
```diff
@@ -170,6 +174,53 @@ def map_years(
...
+            input_data: Input data for datetime mapping. Its index must be pandas.DatetimeIndex.
+            flat: Same as the argument in ``map_years``.
+
+        Returns:
+            Pandas DataFrame filled with Intervals of the calendar's frequency.
+            (also see ``map_year`` and ``map_years``)
+        """
+        # check the datetime order of input data
+        first_timestamp = input_data.index[0]
+        last_timestamp = input_data.index[-1]
```
BSchilperoort, line 196: I was wondering if we could just use `input_data.index.max()` and `input_data.index.min()` here. This gives the maximum and minimum datetime values, and does not care about the order.

geek-yang: Thanks for the comment! Indeed I just checked the function and the way you suggest basically can work with input data with backward datetime index as well. Super cool idea!

s2s/time.py
geek-yang: Corrected! Well spot!

s2s/time.py
```diff
@@ -170,6 +174,50 @@ def map_years(
 
         return index
 
+    def map_to_data(self, input_data: Union[pd.Series, pd.DataFrame, xr.Dataset, xr.DataArray],
+        flat: bool = False
+        ) -> pd.DataFrame:
+        """Map the calendar to input data period.
+        
+        Get datetime range from input data and generate corresponding interval index. This method
+        guarentess that the generated interval (calendar) indices would be covered by the input
```
BSchilperoort, line 183: missed this typo last time, sorry! (`guarantees`)

BSchilperoort (approved):
Only saw a small typo, otherwise everything looks good.

s2s/time.py
geek-yang: Sorry @BSchilperoort . My bad, I will correct it. Thanks for spotting!

Peter9192:
Closing this in favour of #22

Issue #19, opened 2022-06-06
Implementing the lag shifting (discard) method


geek-yang:
The lag shifting method will be implemented in the `time.py` module. When initializing the `AdventCalendar` object, a calendar with counting backward time intervals is created. The user need to provide the target time range and the maximum lag value to finalize the calendar by "trimming" those intervals.

For reference, the relevant legacy codes are:
https://github.com/AI4S2S/proto/blob/0dd397f6101573544f10a55a4205e037e69acdfe/forecasting/func_models.py#L153

Our previous discussion about the design of this module can be found in https://github.com/AI4S2S/proto/issues/3.

geek-yang:
This issue is not valid anymore based on the current design. It will be closed.

PR #20, opened 2022-06-07
Implement lag shift trim (discard) function

geek-yang:
As a user, I want to obtain the intervals with preferred time lag for my target period.

This PR closes issue #19 .


geek-yang:
This PR involves too much old design which has been replaced by new implementations. In order to keep the codebase clean, I will close this PR and implement the `discard` function based on the latest codebase.

PR #22, opened 2022-06-10
Add map to data method

### LINKED ISSUE #27: Order of execution regarding `map_to_data` and `discard` methods
In our current design, when calling `map_to_data` we must know the `max_lag` (and call the `discard` function to trim the `AdventCalendar`) to enforce that the input will be entirely covered, like what is shown in the figure below.

However, when calling the `discard` function, the user needs to know how many time lags does he/she have, by looking the `AdventCalendar`. This results in a circular dependency.

We need to define a workflow for the user to solve this problem.

geek-yang:
As a user, I want to generate a calendar with preferred intervals and the time range should cover my data.

This PR closes #27 .


s2s/time.py
```diff
@@ -170,6 +173,48 @@ def map_years(
...
+        first_timestamp = input_data.index.min()
+        last_timestamp = input_data.index.max()
+
+        map_last_year = last_timestamp.year
+        # ensure that the input data could always cover the advent calendar
+        map_first_year = first_timestamp.year + 1
+        # check if the last date(time) is covered by the advent calendar
+        anchor_date_with_year = pd.Timestamp(year=map_last_year, month=self.month, day=self.day)
+        if anchor_date_with_year > last_timestamp:
+            map_last_year = map_last_year - 1
```
Peter9192, line 201: This looks good

s2s/time.py
```diff
@@ -170,6 +173,48 @@ def map_years(
...
+            Pandas DataFrame filled with Intervals of the calendar's frequency.
+            (also see ``map_year`` and ``map_years``)
+        """
+        # check the datetime order of input data
+        first_timestamp = input_data.index.min()
+        last_timestamp = input_data.index.max()
+
+        map_last_year = last_timestamp.year
+        # ensure that the input data could always cover the advent calendar
+        map_first_year = first_timestamp.year + 1
```
Peter9192, line 197: Consider e.g. anchor date 31 Dec and freq 7 days. Then adding that year is not necessary, and would ideally be avoided. Can you check it in a similar way as for the last date?

s2s/time.py
```diff
@@ -170,6 +173,48 @@ def map_years(
...
+        if map_last_year > map_first_year:
+            intervals = self.map_years(map_first_year, map_last_year, flat)
+            # check if the input data cover all the intervals
+            if intervals.iloc[-1,-1].left < first_timestamp:
+                warnings.warn("The last few intervals are not fully covered by the input data.", UserWarning)
+        elif map_last_year == map_first_year:
+            intervals = self.map_year(map_last_year)
+            # check if the input data cover all the intervals
+            if intervals.iloc[-1].left < first_timestamp:
+                warnings.warn("The last few intervals are not fully covered by the input data.", UserWarning)
```
Peter9192, line 212: This line of code is repeated and can be taken out of the if block. Also, not sure if this is still needed, since you now enforce that the input will be entirely covered

tests/test_time.py
```diff
@@ -58,6 +58,58 @@ def test_map_years(self):
         )
         assert np.array_equal(years, expected)
 
+    def test_map_to_data(self):
+        # test the edge value when the input could not cover the anchor date
+        cal = AdventCalendar(anchor_date=(10, 15), freq='180d')
+        # single year covered
+        time_index = pd.date_range('20191020', '20211001', freq='60d')
```
Peter9192, line 65: Tests look good. I think it's good practice to make one test function per test, now you have multiple test cases in one function. I'd recommend splitting them up.

Peter9192 (commented):
Well done, this looks very clean! I still have a few suggestions.

s2s/time.py
geek-yang: Thanks for spotting this! I added a "if" check for the `map_first_year`:

```py
# ensure that the input data could always cover the advent calendar
if self.map_year(first_timestamp.year).iloc[-1].left > first_timestamp:
    map_first_year = first_timestamp.year
else:
    map_first_year = first_timestamp.year + 1
```

Now the case you mentioned here is covered by the function. I put it in the test as well:
```py
    def test_map_to_data_single_year_coverage(self):
        # test the single year coverage
        cal = AdventCalendar(anchor_date=(12, 31), freq='180d')
        # multiple years covered
        time_index = pd.date_range('20210101', '20211231', freq='7d')
        test_data = np.random.random(len(time_index))
        timeseries = pd.Series(test_data, index=time_index)
        year = cal.map_to_data(timeseries)

        expected = np.array(
            [
                interval("2021-07-04", "2021-12-31"),
                interval("2021-01-05", "2021-07-04"),
            ]
        )

        assert np.array_equal(year, expected)
```

tests/test_time.py
geek-yang: Thanks for the comment. I split those test cases now and rename each test descriptively.

s2s/time.py
geek-yang: If we take the code out of this block, then we need to make extra check for the dimension of intervals to refer to the last element (advent calendar can be 2d or flat in 1d), which is not so nice. I prefer to keep it this way for now. It is needed only temporarily. When we implement the `discard` function with `max_lag` and trim the calendar, we know exactly what the calendar looks like and can then enforce that the input will be entirely covered (currently the incomplete intervals are allowed, e.g. the `AdventCalendar` goes with `archor date = (10, 15)` and `freq='30d'`, the input data is from 2021.11.10 - 2022.12.15, 10 days frequency, then the last interval "(2020-10-20, 2020-11-19]" is slightly incomplete, but we need to warn the user).

This also leads to another issue #27 , about the order of calling these methods (e.g. do we expect our user to instantiate `AdventCalendar` first, map to data, then discard intervals based on `max_lag`?). I have discussed it with @BSchilperoort and opened an issue for this, let's bring our discussion there.

Peter9192: In that case, can you write a test that triggers this warning?

Peter9192: Reviewing this from my phone, so couldn't test it, but shouldn't the iloc be 0 instead of -1?

geek-yang: Based on the discussion in #27 , this problem is solved then. We don't need those warnings now as we ensure that the input will be entirely covered.

Peter9192 (approved):
Great, just one more small suggestion. Thanks for seeing this through!

s2s/time.py
```diff
@@ -192,37 +192,33 @@ def map_to_data(self, input_data: Union[pd.Series, pd.DataFrame, xr.Dataset, xr.
...
 
         return intervals
 
     def __str__(self):
-        return f"{self.n} periods of {self.freq} leading up to {self.month}/{self.day}."
+        return f"{self._n_intervals} periods of {self.freq} leading up to {self.month}/{self.day}."
 
     def __repr__(self):
-        props = ", ".join([f"{k}={v}" for k, v in self.__dict__.items()])
+        props = ", ".join([f"{k}={v}" for k, v in self.__dict__.items() if '_' not in k])
```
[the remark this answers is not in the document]
geek-yang: Indeed, that's also what I think. Thanks for the refactoring. I will commit it.

PR #23, opened 2022-06-13
Initial implementation of traintest methods

Peter9192:
work in progress

Todo:
- [x] overwrite if method is different?
- [x] need to store self._df internally after calling map_years/map_data
- [x] implement tests
- [x] add first (simple) method


geek-yang:
@Peter9192 This PR is ready for review now. Note that the failed test has nothing to do with this PR (due to the space in docstrings).

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+        """
         raise NotImplementedError
 
-    def get_train_test_indices(self, strategy, params):  # noqa
-        """Shorthand for getting both train and test indices."""
-        # train = self.get_train_indices()
-        # test = self.get_test_indices()
+    def get_test(self) -> pd.DataFrame:
+        """Return indices for test data indices using given strategy.
+        """
```
Peter9192, line 525: Why are these still not implemented?

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+            1         2021            1  (2020-10-20, 2021-04-18]  train   test
+            2         2020            0  (2020-04-18, 2020-10-15]   test  train
+            3         2020            1  (2019-10-21, 2020-04-18]   test  train
+        """
+        # checker if generated intervals are flat.
+        if self.intervals.ndim != 1:
+            raise ValueError("Please set `flat = True` when calling `map_years` or `map_data`")
+
+        # infer anchor years from give intervals
+        anchor_years = [timestamp.right.year for timestamp in self.intervals[::self._n_intervals]]
```
Peter9192, line 488: If you store self.intervals in 2d, you can just get the anchor years from the row index, right?

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+        # Create DataFrame from intervals Series to store the train/test groups
+        traintest_base = pd.DataFrame(data = {
+        'anchor_year': np.repeat(anchor_years, self._n_intervals),
+        'i_intervals': np.tile(range(self._n_intervals), len(anchor_years)),
+        'intervals': self.intervals,
+        })
```
Peter9192, line 494: This looks to be misaligned

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+
+        # checker if the method is configured.
+        if self._traintest_method is not None:
+            self.traintest = traintest.ALL_METHODS[self._traintest_method](traintest_base,
+                **self._method_kwargs)
```
Peter9192, line 499: Raise a helpful error message when the method has not been set yet.

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+
+        return self.traintest
+
+    def set_traintest_method(self, method: str, **method_kwargs: Optional[dict]):
+        """
+        The user must choose a method here. And the method will be used by
+        all traintest splitting methods.
```
Peter9192, line 506: "the method will be used by all methods"? This is not accurate.

```suggestion
        Configure the train/test splitting strategy for this calendar instance.
```

s2s/traintest.py
```diff
@@ -0,0 +1,62 @@
...
+    consecutive folds.
+
+    For more information
+
+    args:
+        intervals_df: DataFrame to store the train/test groups.
+        n_splits: number of train/test splits.
+        # To do
+        Gap_prior: skip n groups/years prior to every test group/year
+        Gap_after: skip n groups/years after every test group/year
```
Peter9192, line 22:
```suggestion
```
Not used, so leave them out of public API for now

s2s/traintest.py
```diff
@@ -0,0 +1,62 @@
...
+
+def kfold(intervals_df: pd.DataFrame, n_splits: int = 2, **kwargs):
+    """
+    K-Folds cross-validator, which splits provided intervals into k
+    consecutive folds.
+
+    For more information
+
+    args:
+        intervals_df: DataFrame to store the train/test groups.
```
Peter9192, line 18:
```suggestion
        df: DataFrame to store the train/test groups. It must have a column named "anchor_years"
```

1. "intervals_df" is mostly meaningful in the context of our Calendar object, whereas this function is much more generic, so I would use a more generic term.
2. We could consider making the column name "anchor_year" also settable by the user, so they could reuse the function independently from the calendar. 
3. Ideally, I would want anchor_year (or whatever other thing) to be the index. In that case, this function reduces to mostly bare sklearn, which is good. I'll see if I can come up with a proposal for that.

tests/test_time.py
```diff
@@ -281,3 +254,46 @@ def test_non_time_dim(self, dummy_calendar, dummy_dataset):
...
+
+    def test_set_traintest_method(self, dummy_calendar):
+        dummy_calendar.set_traintest_method("kfold", n_splits = 2)
+        assert dummy_calendar._traintest_method == "kfold" # pylint: disable=protected-access
+        assert dummy_calendar._method_kwargs["n_splits"] == 2 # pylint: disable=protected-access
+
+    def test_set_traintest_method_not_support(self, dummy_calendar):
+        # test when the given method is not supported
+        with pytest.raises(ValueError):
+            dummy_calendar.set_traintest_method("not_a_real_method")
```
Peter9192, line 272: Add a test for "traintest method not set"?

tests/test_time.py
```diff
@@ -281,3 +254,46 @@ def test_non_time_dim(self, dummy_calendar, dummy_dataset):
...
+
+    def test_get_traintest(self, dummy_calendar):
+        dummy_calendar.map_years(2019, 2021, flat=True)
+        dummy_calendar.set_traintest_method("kfold", n_splits = 2)
+        traintest_group = dummy_calendar.get_traintest()
+        # check the first fold
+        expected_group = ['train', 'train', 'test', 'test', 'test', 'test']
```
Peter9192, line 279: Works because the default option for sklearn is set to `shuffle=False`, otherwise groups would have been random.

tests/test_time.py
```diff
@@ -281,3 +254,46 @@ def test_non_time_dim(self, dummy_calendar, dummy_dataset):
...
+        assert np.array_equal(traintest_group["fold_0"].values, expected_group)
+    
+    def test_get_traintest_intervals_not_flat(self, dummy_calendar):
+        dummy_calendar.map_years(2019, 2021, flat=False)
+        dummy_calendar.set_traintest_method("kfold", n_splits = 2)
+        # test when the generated intervals are not flat
+        with pytest.raises(ValueError):
+            dummy_calendar.get_traintest()
```
Peter9192, line 287: I think this needs to be reversed somehow.

tests/test_traintest.py
```diff
@@ -0,0 +1,41 @@
...
+        # create calendar object
+        cal = AdventCalendar(anchor_date=(10, 15), freq=freq)
+        intervals = cal.map_years(2019, 2021, flat=True)
+        # infer anchor years from give intervals
+        anchor_years = [timestamp.right.year for timestamp in intervals[::n_intervals]]
+        # Create DataFrame from intervals Series to store the train/test groups
+        traintest_base = pd.DataFrame(data = {
+        'anchor_year': np.repeat(anchor_years, n_intervals),
+        'i_intervals': np.tile(range(n_intervals), len(anchor_years)),
+        'intervals': intervals,
```
Peter9192, line 24:
```suggestion
    def dummy_intervals_df(self):
        freq = "180d"
        n_intervals = pd.Timedelta("365days") // pd.to_timedelta(freq)
        # create calendar object
        cal = AdventCalendar(anchor_date=(10, 15), freq=freq)
        intervals = cal.map_years(2019, 2021, flat=True)
        # infer anchor years from give intervals
        anchor_years = [timestamp.right.year for timestamp in intervals[::n_intervals]]
        # Create DataFrame from intervals Series to store the train/test groups
        traintest_base = pd.DataFrame(data = {
        'anchor_year': np.repeat(anchor_years, n_intervals),
        'i_intervals': np.tile(range(n_intervals), len(anchor_years)),
        'intervals': intervals,
```

This is still very specific to the calendar use case, whereas I think in general this function can be much more generic. Could we just generate a random dataframe instead?

Peter9192 (commented):
Can't approve or request changes since I originally opened this PR.

I think it's a good start, but it can still be done neater. My two most important concerns:

1. I think we should store the 2d (not flat) representation of intervals_df internally, and also pass this one to the traintest methods. That way, the train test methods will be much simpler, perhaps even redundant (at this stage)
2. The traintest methods (in traintest.py) are now very much tailored to the use case in which they work in tandem with the calendar. I think they can be made more generic, so they can also be used more independently.

I'll make a suggestion as to how we could make it work with the 2d version of df_intervals.

s2s/time.py
```diff
@@ -48,16 +48,18 @@
...
+import numpy as np
 import pandas as pd
 import xarray as xr
+from s2s import traintest
 
 
 PandasData = (pd.Series, pd.DataFrame)
 XArrayData = (xr.DataArray, xr.Dataset)
 
-
```
Peter9192, line 60: Linter says: two spaces around class definitions

s2s/time.py
Peter9192: Ah, I guess it's ambiguous what to return, e.g. for K-fold when you have multiple folds. Need to discuss what behavious we expect.

s2s/traintest.py
```diff
@@ -0,0 +1,62 @@
...
+
+    for i, (train_index, test_index) in enumerate(folds):
+        # get train year index and test year index
+        train_years = [anchor_years[index] for index in train_index]
+        test_years = [anchor_years[index] for index in test_index]
+        # label train/test group
+        col_name = 'fold_' + str(i)
+        intervals_df[col_name] = 'skip'
+        intervals_df.loc[intervals_df['anchor_year'].isin(train_years), col_name] = 'train'
+        intervals_df.loc[intervals_df['anchor_year'].isin(test_years), col_name] = 'test'
```
Peter9192, line 43: Alternatively, we might consider using GroupKFold: https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GroupKFold.html

s2s/time.py
```diff
@@ -48,16 +48,18 @@
 
 """
 import warnings
+from typing import Optional
 from typing import Tuple
 from typing import Union
```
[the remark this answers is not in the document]
Peter9192: Perhaps we already do.. Anyway, I had expected isort to combine these..

s2s/traintest.py
```diff
@@ -1,6 +1,6 @@
 """AI4S2S train/test splitting methods.
 
-A collection of train/test splitting approaches.
+A collection of train/test splitting approaches used for cross-validation.
```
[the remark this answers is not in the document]
geek-yang: Corrected.

s2s/time.py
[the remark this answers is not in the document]
geek-yang: If we put them together, the linter will complain. That's why 😄 .

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
-        if start and end:
-            pass
-        elif start and periods:
+        if start is not None:
             pass
-        elif end and periods:
+        elif periods:
             pass
         else:
             raise ValueError("Of start/end/periods, specify exactly 2")
```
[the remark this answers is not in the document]
geek-yang: Corrected.

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
 
-    def get_train_indices(self, strategy, params):  # noqa
-        """Return indices for training data indices using given strategy."""
+    def get_cv_groups(self):
+        """Group intervals into bins.
+
+        Group intervals into bins for each anchor year by labelling.
+        args:
+            intervals: output of ``map_years``.
+        """
```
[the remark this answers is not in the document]
geek-yang: This can be removed. Thanks for spotting this.

s2s/time.py
geek-yang: Yes, that's exactly what I was thinking. 😉

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+        The user must choose a method here. And the method will be used by
+        all traintest splitting methods.
+
+        Args:
+            method: one of the methods available in `s2s.traintest`
+            method_kwargs: keyword arguments that will be passed to `method`
+        """
+        # checker if the given method is supported in `s2s.traintest`
+        if method not in [key for key, _ in traintest.ALL_METHODS.items()]:
```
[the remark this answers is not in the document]
geek-yang: Solved by `func = traintest.ALL_METHODS.get(method)`.

s2s/time.py
```diff
@@ -417,38 +421,106 @@ def __repr__(self):
...
+            >>> calendar.get_traintest()
+               anchor_year  i_intervals                 intervals fold_0 fold_1
+            0         2021            0  (2021-04-18, 2021-10-15]  train   test
+            1         2021            1  (2020-10-20, 2021-04-18]  train   test
+            2         2020            0  (2020-04-18, 2020-10-15]   test  train
+            3         2020            1  (2019-10-21, 2020-04-18]   test  train
+        """
+        # checker if generated intervals are flat.
+        if self.intervals.ndim != 1:
+            raise ValueError("Please set `flat = True` when calling `map_years` or `map_data`")
```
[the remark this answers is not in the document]
geek-yang: Solved by adding a `flat` property to the generated intervals.

s2s/time.py
geek-yang: Right. Now we always save `self._intervals` in 2D.

s2s/time.py
geek-yang: Not needed anymore.

s2s/traintest.py
geek-yang: Thanks for the suggestions. I totally agree! And by having the `anchor_year` as the index and attach train/test group information to it, then we can extract the train and test group separately by simply calling `pandas.DataFrame.groupby()`.

geek-yang: The docstrings have been updated.

tests/test_time.py
geek-yang: Yes, I'm aware of it. But maybe not clear for the user/developer. I will add this argument in the function.

tests/test_time.py
geek-yang: This function is not needed anymore.

tests/test_traintest.py
geek-yang: N/A with the latest updates.

s2s/traintest.py
```diff
@@ -0,0 +1,62 @@
...
+        intervals_df.loc[intervals_df['anchor_year'].isin(train_years), col_name] = 'train'
+        intervals_df.loc[intervals_df['anchor_year'].isin(test_years), col_name] = 'test'
+
+    return intervals_df
+
+def random_strat():
+    raise NotImplementedError
+
+def timeseries_split():
+    raise NotImplementedError
```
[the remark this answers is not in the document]
geek-yang: Corrected! (Just call "Black" formatter).

s2s/traintest.py
geek-yang: Thanks for the suggestion. I think "KFold" provides more control (e.g. shuffle and random seed) than this one. I would prefer to keep the "KFold" implementation.

s2s/time.py
```diff
@@ -414,41 +412,108 @@ def __str__(self):
...
+            [f"{k}={v}" for k, v in self.__dict__.items() if not k.startswith("_")]
+        )
         return f"AdventCalendar({props})"
 
-    def discard(self, max_lag):
+    @property
+    def flat(self):
+        if self._intervals is not None:
+            return self._intervals.stack()
+        raise ValueError("The calendar is not initialized with intervals yet.")
```
Peter9192, line 437: Still to do: clarify this error/convert to warning by saying something like: "use calendar.map_years or calendar.map_to_data"

s2s/time.py
```diff
@@ -414,41 +412,108 @@ def __str__(self):
...
+            >>> calendar.set_traintest_method("kfold", n_splits = 2)
+            >>> traintest_group = calendar.traintest
+            >>> traintest_group
+                                                              0                         1
+            anchor_year fold_0 fold_1
+            2021        test   train   (2021-04-18, 2021-10-15]  (2020-10-20, 2021-04-18]
+            2020        train  test    (2020-04-18, 2020-10-15]  (2019-10-21, 2020-04-18]
+        """
+        if self._traintest is None:
+            raise RuntimeError("The train/test splitting method has not been set yet.")
```
Peter9192, line 492: Error message can be more helpful by hinting at how to set it

s2s/time.py
```diff
@@ -414,41 +412,108 @@ def __str__(self):
...
-        raise NotImplementedError
+    @property
+    def traintest(self) -> pd.DataFrame:
+        """Shorthand for getting both train and test indices.
 
-    def get_test_indices(self, strategy, params):  # noqa
-        """Return indices for test data indices using given strategy."""
-        raise NotImplementedError
+        The user will get a flat intervals list similar to the output of `map_years`
+        or `map_data`, but with extra columns containing `train` and `test` labels.
```
Peter9192, line 465: Document that the user must first set the method (using calendar.set_traintest_method)

s2s/time.py
```diff
@@ -414,41 +412,108 @@ def __str__(self):
...
+        df_combined = self._intervals.join(self._traintest)
+        new_index_cols = [self._intervals.index.name] + list(self._traintest.columns)
+        return df_combined.reset_index().set_index(new_index_cols)  # .stack()
+
+    def set_traintest_method(
+        self, method: str, overwrite: bool = False, **method_kwargs: Optional[dict]
+    ):
+        """
+        Configure the train/test splitting strategy for this calendar instance.
+
```
Peter9192, line 502: Can we add a hyperlink to the documentation for s2s.traintest that will render on readthedocs?

Peter9192 (commented):
Well done! I had a quick look and I think we should probably just merge this. Core functionality looks well designed like this. Future PRs can address any issues or additional ideas with respect to this functionality.

Can't approve since I originally opened this PR

README.md
```diff
@@ -2,12 +2,12 @@
 
 ## Badges
 
-[![github repo badge](https://img.shields.io/badge/github-repo-000.svg?logo=github&labelColor=gray&color=blue)](https://github.com/AI4S2S/ai4s2s)
+[![github repo badge](https://img.shields.io/badge/github-repo-000.svg?logo=github&labelColor=gray&color=blue)](https://github.com/AI4S2S/s2spy)
```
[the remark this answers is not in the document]
geek-yang: Indeed, otherwise the test complains...very annoying...

s2s/time.py
geek-yang: "use calendar.map_years or calendar.map_to_data" is added to the error message now.

s2s/time.py
geek-yang: Add "Use `set_traintest_method` to set the method." to the error message.

s2s/time.py
geek-yang: Add "The user must first set a method using `set_traintest_method`." to the docstrings.

s2s/time.py
geek-yang: https://ai4s2s.readthedocs.io/en/latest/autoapi/s2spy/time/index.html
This link will be added to the docstring, to tell the user the where to find a list of train/test splitting methods supported by this function.

geek-yang (approved):
Approved. I will merge it.

geek-yang:
> Well done! I had a quick look and I think we should probably just merge this. Core functionality looks well designed like this. Future PRs can address any issues or additional ideas with respect to this functionality.
> 
> Can't approve since I originally opened this PR

Thanks a lot for the review. I just addressed all the comments and merge the changes from #31. I will go ahead and merge this PR. Thanks again for the quick response. 😆

PR #31, opened 2022-06-28
Modifications for publication on PyPI 

BSchilperoort:


.bumpversion.cfg
```diff
@@ -1,10 +1,10 @@
 [bumpversion]
-current_version = 0.1.0
```
Peter9192, line 2: Why go from 1.0 to 0.1? If the template starts from 0.1.0, I think that makes sense. This is not a bugfix, we started adding features right away.

s2spy/__init__.py
```diff
@@ -1,9 +1,9 @@
...
+"""Documentation about s2spy"""
 import logging
 
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 __author__ = "Yang Liu"
 __email__ = "y.liu@esciencecenter.nl"
-__version__ = "0.1.0"
+__version__ = "0.0.1"
```
Peter9192, line 9: same

setup.cfg
```diff
@@ -21,11 +21,11 @@ keywords =
     ai
 long_description = file: README.md
 long_description_content_type = text/markdown
-name = s2s
+name = s2spy
 project_urls =
     Bug Tracker = https://github.com/AI4S2S/ai4s2s/issues
 url = https://github.com/AI4S2S/ai4s2s
-version = 0.1.0
+version = 0.0.1
```
Peter9192, line 28: also

Peter9192 (approved):
Very superficial review, my only comment is the version number. I'd stick with 0.1.0 unless there's a very good reason not to (and then prefer to move to 0.1.1)

.bumpversion.cfg
BSchilperoort: Ok! I'll keep it at 0.1.0.

PR #36, opened 2022-06-29
Implement minimum example of RGDR (dimensionality reduction) usecase

geek-yang:
Only for prototyping the response guided dimensionality reduction (RGDR), which is the core functionality in the module of dimensionality reduction, DO NOT MERGE!

This PR relates to the discussion in #38 .


BSchilperoort:
Implemented by #57

PR #37, opened 2022-06-30
Some suggestions for an alternative implementation of traintest

Peter9192:
Discussion points:
- What to do with 'flat' keyword?
- Executing train/test only once
- Do as little as possible in 'traintest.py'?
- ...

How to use:

```py
import s2s.time
cal = s2s.time.AdventCalendar((1, 31))
cal.map_years(2019, 2022)
cal.set_traintest_method('kfold', n_splits=2)
cal.traintest
```

### DIFF
--- a/s2s/time.py
+++ b/s2s/time.py
@@ -37,7 +37,7 @@ Example:
     2020         (2020-07-04, 2020-12-31]  (2020-01-06, 2020-07-04]
 
     >>> # To get a stacked representation:
-    >>> calendar.map_years(2020, 2022, flat=True)
+    >>> calendar.map_years(2020, 2022).flat
     0    (2022-07-04, 2022-12-31]
     1    (2022-01-05, 2022-07-04]
     2    (2021-07-04, 2021-12-31]
@@ -95,7 +95,7 @@ class AdventCalendar:
         self._n_intervals = pd.Timedelta("365days") // pd.to_timedelta(freq)
         self._n_target = 1
         self._traintest = None
-        self.intervals = None
+        self._intervals = None
 
     def _map_year(self, year: int) -> pd.Series:
         """Internal routine to return a concrete IntervalIndex for the given year.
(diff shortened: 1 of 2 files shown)


Peter9192:
Todo:

- [ ] Fix tests
- [ ] Update docstrings

Feel free to merge into your own feature branch and continue working on it there.

s2s/time.py
```diff
@@ -418,9 +410,18 @@ def __str__(self):
         return f"{self._n_intervals} periods of {self.freq} leading up to {self.month}/{self.day}."
 
     def __repr__(self):
+        if self._intervals is not None:
+            return str(self._intervals)
```
geek-yang, line 414: In notebook the calendar table looks a bit weird after mapping years to the calendar due to this brute-force conversion:

```py
i_interval                         0                         1   \
anchor_year                                                       
2021         (2021-09-15, 2021-10-15]  (2021-08-16, 2021-09-15]   
2020         (2020-09-15, 2020-10-15]  (2020-08-16, 2020-09-15]   
2019         (2019-09-15, 2019-10-15]  (2019-08-16, 2019-09-15]   

i_interval                         2                         3   \
anchor_year                                                       
2021         (2021-07-17, 2021-08-16]  (2021-06-17, 2021-07-17]   
2020         (2020-07-17, 2020-08-16]  (2020-06-17, 2020-07-17]   
2019         (2019-07-17, 2019-08-16]  (2019-06-17, 2019-07-17] 
```
Actually `__repr__` is supposed to return strings in practice, rather than a table. I think we'd better not use it this way. Since the 2D table view of the calendar is always a good feedback to the user, it is ok to make it visible `self.intervals`.

geek-yang (approved):
Thanks for the suggestions! The code looks quite light-weight now. There is only one minor issue related to the display of calendar in the notebook. I will address it in PR #23 after merging this PR to it.

s2s/time.py
Peter9192: Ah, my bad! We should return the __repr__ method of the pandas dataframe instead

Peter9192:
```suggestion
        if self._intervals is not None:
            return repr(self._intervals)
```

Doest that make it better?

geek-yang: 😓 Not really...It still looks the same. I think internally `repr(object)` tries to convert things to strings. The display of dataframe looks ok in jupyter notebook because it is treated as a html object (I learn this from google...). So the conversion to strings does not work.

geek-yang: Fixed by adding a `def _repr_html_(self):` function specifically for jupyter notebooks.

The solution is found [here](https://stackoverflow.com/questions/32848845/what-pandas-dataframe-method-tells-ipython-notebook-to-display-as-html).

PR #43, opened 2022-07-12
Implemented max_lag

BSchilperoort:
- [x] implement max_lag kwarg
- [x] allow max_lag that exceeds 365 days, automatically skip anchor years
- [x] write tests for new functionality

### DIFF
--- a/s2spy/time.py
+++ b/s2spy/time.py
@@ -67,7 +67,7 @@ class AdventCalendar:
 
     def __init__(
         self, anchor_date: Tuple[int, int] = (11, 30), freq: str = "7d",
-        max_lag = None
+        max_lag: int = None
     ) -> None:
         """Instantiate a basic calendar with minimal configuration.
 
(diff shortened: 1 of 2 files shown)


tests/test_time.py
```diff
@@ -7,6 +7,9 @@
 from s2spy.time import AdventCalendar
 
 
+# pylint: disable=protected-access
```
geek-yang, line 10: This is really nice!

s2spy/time.py
```diff
@@ -93,11 +101,20 @@ def __init__(
...
         self._n_target = 1
         self._traintest = None
         self._intervals = None
 
+        periods_per_year = pd.Timedelta("365days") / pd.to_timedelta(freq)
+        # Determine the amount of intervals, and number of anchor years to skip
+        if max_lag:
+            self._n_intervals = max_lag + self._n_target
+            self._skip_years = np.ceil(self._n_intervals /
+                                       periods_per_year).astype(int) - 1
```
geek-yang, line 113: Nice design! In this way we avoid train/test information leakage elegantly.

geek-yang (approved):
Nice work! It is nice that by this implementation one can avoid train/test information leakage through skipping years. In the future we would have usecases that allows the overlapping of years (it is described here #44 ). But we can accommodate that easily by providing another key word argument for to enable the option. Well done 😄 !

Issue #38, opened 2022-06-30
Create RGDR modules


semvijverberg:
We want to have a minimal workflow to test our pipeline, we like to implement the RGDR method to reduce dimensionality. The approach broadly consists of the following steps, of which a basic minimum workflow is not implemented in prototype_RGDR.ipynb in branch (prototype_RGDR_usecase). For more info on the method, see [this paper](https://www.nature.com/articles/s41612-022-00237-7).

1. gridpoint-wise correlation map analysis versus 1-d target
2. clusters these region via DBSCAN
3. calculate 1-d timeseries based on each label (i.e. precursor region)

My vision is create two separate python files:
1. one purely for the (partial) correlation map analysis , with also support for regression analysis; now called map_analysis.py 
2. the second for the clustering analysis (people might want to use this functionality stand-alone; now called cluster_regions.py

I already created some NotImplemented functions in both python files. To a large extent, the cluster_regions.py is a refractor of the original find_precursors.py in proto.

geek-yang:
Thanks for the explanation and for completing the minimum example @semvijverberg . This #36 already seems to be a very nice starting point. Now, I think we need to discuss the design of the new module. Going back to the whiteboard discussion we had before, my envision about the modules in the package looks like this:

![image](https://user-images.githubusercontent.com/25911757/176748075-bdf0996d-732e-435b-b6a7-6e668c074523.png)

Basically, what we are trying to address here, is the module for dimensionality reduction. In this module, we want to have some methods, like:
- PCA
- MCA
- RGDR (Response guided dimensionality reduction)
- ...

For most of the methods, the implementations are not difficult and there are packages we could make use of easily (e.g. statsmodel, scipy, scikit-learn). Only for RGDR, we need to code ourselves. But this method is quite valuable (and essential for s2s) and we want to include it as a minimum example.

This is my understand about what we've done so far. Before diving into the implementation of RGDR method, we need to think about the general API design of this module, and how do we want this module to work with the module we already have (e.g. time.py, traintest.py), and the module we skip for now but would like to have later (e.g. preprocess module). When we decide the design of this module, we will know what python files do we need and how will we organize them. For now, RGDR is just one method (although it carries the largest work-load for us), let's think about the overall design first and then get into details about the implementation.

> My vision is create two separate python files:
> one purely for the (partial) correlation map analysis , with also support for regression analysis; now called map_analysis.py
[...]

PR #49, opened 2022-07-18
Implemented correlation map function

BSchilperoort:
This PR implements the correlation_map function required for RGDR. Related to #38 

Validation of the input data is important, as `xr.apply_ufunc` errors are not very user-friendly.


geek-yang (approved):
Looks really really concise! Nice work! I also verify the functionality of it with the results from old implementation (legacy code in the minimum example). That works pretty well. Follow our discussion, one last thing to change is the name of `time_dim` argument. Afterwards, just go ahead and merge it!

Peter9192:
I just wonder if _rgdr/map_analysis.py is the most logical path/name for this functionality. I might have put it in utils.py or correlation.py for now and see whether it grows into something more generic in the long run.

Issue #14, opened 2022-05-25
Add MyPy


Peter9192:
In #12 I started adding type annotations. However, mypy hasn't yet been set up for this repo, so they haven't been checked yet. It would be good to configure mypy asap, so as to avoid accumulating many type issues

BSchilperoort:
I quickly installed mypy to test the package. Things mostly go OK, however some of our dependencies do not having type hinting.

```
s2spy\traintest.py:8: error: Skipping analyzing "pandas": module is installed, but missing library stubs or py.typed marker
s2spy\traintest.py:10: error: Skipping analyzing "sklearn.model_selection._split": module is installed, but missing library stubs or py.typed marker
```

BSchilperoort:
Currently I get no mypy errors when running mypy locally.
I have the exact same versions of Python (3.10.5) mypy (0.971) and mypy-extensions installed. Local machine is windows.

<hr>

One of the two errors that GH actions shows is the following;
```
s2spy/_resample.py
  Line: 140
    mypy: error / Argument 1 to "rename" of "Dataset" has incompatible type "str"; 
                  expected "Optional[Mapping[Any, Hashable]]" (col 48)
```

Which is strange because this is within an if isinstance statement;
```python
139    if isinstance(input_data, xr.DataArray) and interval_means.name is None:
140        interval_means = interval_means.rename("mean_values")
```

Peter9192:
Hmm that's curious. [This article](https://adamj.eu/tech/2021/05/17/python-type-hints-how-to-narrow-types-with-isinstance-assert-literal/) suggests that mypy should be able to infer the narrowed type.

BSchilperoort:
I now know why this error shows up. I test for the type of input_data and not of interval_means.

They will have the same type, but mypy is not able to infer this, and the if statement should probably check for the type of interval_means and not input_data.

However, mypy is still not giving me any errors locally. When using `reveal_type()` in the code, mypy does however print this to the terminal:
```
>>> mypy . --ignore-missing-imports --exclude=build

s2spy\_resample.py:141: note: Revealed type is "xarray.core.dataarray.DataArray"
Success: no issues found in 16 source files
```

Peter9192:
Merged as part of #60

PR #53, opened 2022-07-22
Refactor traintest to support sklearn splitters directly

### LINKED ISSUE #52: Check availability of legacy code train/test splitting methods
The new train/test implementation relies on sklearn for the splitter classes. These mostly correspond to the legacy code methods (see image below)

image

@semvijverberg could you describe here what the `leave` splitter is supposed to do?

BSchilperoort:
Originally we intended to write wrappers for sklearn's splitter classes, and use these to split the data, e.g.:
```py
split_data = kfold(data)
```

However, this would force us to re-implement or write wrappers for every single sklearn splitter. This PR implements code where the splitters are directly supported in a single wrapper function, e.g.
```py
from sklearn.model_selection import KFold
split_data = fold_by_group(data, group='anchor_year', splitter=KFold())
```
With this change we directly support every single splitter sklearn currently has implemented, and even ones they might implement in the future, all while having very little code to maintain in s2spy.

Closes #52


s2spy/traintest.py
```diff
@@ -2,52 +2,94 @@
...
-    for i, (train_index, test_index) in enumerate(folds):
-        col_name = f"fold_{i}"
-        output[col_name] = "skip"
-        output[col_name].iloc[train_index] = "train"
-        output[col_name].iloc[test_index] = "test"
+    if isinstance(data, xr.Dataset):
+        data = data.copy()  # otherwise the dataset is modified inplace
 
-    return output
+        assert (
```
Peter9192, line 36: I'm not sure where I got it from, but I always thought it bad practice to use assert statements outside of tests. See e.g. here for some relevant discussion: https://community.sonarsource.com/t/feature-python-assert-should-be-consider-harmful/38501/3.

Some thoughts:
+ Assertions can be short and simple
- Assertions can be removed in optimized production code
- Assertions always raise assertionerrors as opposed to a more befitting error message.

An alternative and equally short solution would be:

```py
if len(data.anchor_year.size) < 1:
    raise InputError("Input data must have more than 1 anchor year to split")
```

with the only upside/downside that you can/need to define the `InputError` (or you can be lazy and use a ValueError instead)

s2spy/traintest.py
```diff
@@ -2,52 +2,94 @@
...
+            indices[test_indices] = "test"
+            train_test_dict = dict(zip(anchor_years, indices))
+
+            # map anchor_year row values with train/test
+            data[col_name] = data["anchor_year"].map(train_test_dict)
+
+    return data
+
+
+def leave_n_out():
```
Peter9192, line 84: Input for a later discussion: should this then be a separate function or an option of the above?

s2spy/traintest.py
```diff
@@ -2,52 +2,94 @@
...
+    """cross validator that leaves out n anchor years, instead of cv
+    defined on number of splits
+    """
+    # yet to be implemented
     raise NotImplementedError
 
 
-def iter_traintest(traintest_group, data, dim_function = None):
+def iter_traintest(traintest_group, data, dim_function=None):
```
Peter9192, line 92: Instead of adding non-functional code to a repo that we'll implement "someday/maybe", it would be better to open an issue to discuss it. Shall we leave it out for now?

s2spy/traintest.py
```diff
@@ -2,52 +2,94 @@
...
+import xarray as xr
 
 
-def kfold(df: pd.DataFrame, n_splits: int = 2, shuffle=False, **kwargs):
-    """
-    K-Folds cross-validator, which splits provided intervals into k
-    consecutive folds.
-
-    For more information
+def fold_by_anchor(cv, data: Union[xr.Dataset, pd.DataFrame]):
```
Peter9192, line 11: I'm not sure that the name "fold_by_anchor" is intuitive. The anchor is quite specific to our use case with the calendar, whereas this wrapper is fairly generic. What about something like

```suggestion
def fold_grouped_data(cv, data: Union[xr.Dataset, pd.DataFrame], key: Optional[Str]="group"):
```

where `group` would be replaced by "anchor_year" for our use case?

Peter9192: Additionally, `cv` is a bit of a confusing name. It's an abbreviation of "crossvalidator" or something akin, but in the docstrings we consistently refer to it as splitter class. Similarly, splitter and fold are used interchangeably. Shall we use split for everything, i.e. `split_by_anchor(splitter, data: ...)`?

Peter9192: Furthermore, as a nice-to-have: I was wondering if it is possible to also give a type hint for the splitter (cv). It seems they all derive from https://github.com/scikit-learn/scikit-learn/blob/baf0ea25d6dd034403370fea552b21a6776bef18/sklearn/model_selection/_split.py#L54, can we use that?

Peter9192: Finally, can we explain how this differs from skearn's existing GroupKFold/GroupShuffleSplit, and why that is useful?

s2spy/traintest.py
```diff
@@ -2,52 +2,94 @@
...
+            (cv.get_n_splits(data.anchor_year.size), data.anchor_year.size), dtype="<U6"
+        )
+
+        for i, (train_indices, test_indices) in enumerate(folds):
+            fold_data[i, :] = "skip"
+            fold_data[i, train_indices] = "train"
+            fold_data[i, test_indices] = "test"
+
+        data.expand_dims("fold")
+        data["fold"] = np.arange(fold_data.shape[0])
```
Peter9192, line 52: also here be consistent with "split" vs "fold". I think some of the methods like kfold are both folds and splits, but others like shufflesplits are splits but not folds.

s2spy/traintest.py
```diff
@@ -2,52 +2,94 @@
...
+        data["traintest"] = (["fold", "anchor_year"], fold_data)
+        data = data.set_coords("traintest")
+
+    else:
+        # split the anchor years
```
Peter9192, line 57:
```suggestion
    else:
        # input data is a pandas dataframe
```

Peter9192: When I read this I am very tempted to split it into three functions:

```Py
def fold_by_anchor(cv, data: Union[xr.Dataset, pd.DataFrame]):
    """This is the public API so full docstring comes here."""
    if isinstance(data, xr.Dataset):
        return _fold_dataset(cv, data)
    elif isinstance(data, pd.DataFrame):
        return _fold_dataframe(cv, data)
    raise TypeError("Input data should be xr.dataset or pd.dataframe")


def _fold_dataset(cv, data: xr.Dataset):
   # implementation here


def _fold_dataframe(cv, data: pd.DataFrame):
   # implementation here

```

That also neatly gets rid of all your assertions.

I was trying to find where I picked this up, but I can't find a good source. Anyway, it relates to best practices such as "a function should do only one thing and do one thing well", "a function should not be too long", and the idea of multiple dispatch (one of the main idioms of the Julia language).

Peter9192 (changes requested):
Awesome work! This is a prime example of "something much simpler and more elegant is hiding underneath this messy spaghetti code". Very nice and clean commit history as well. I would appreciate a bit of rationale in the top post of the PR, or have it linked to a corresponding issue in which the rationale is explained. Seeing that this function is much more generic than the old version and could potentially be of interest to many users, I think it would help to think about more intuitive and more generic function names. Made some suggestions.

s2spy/traintest.py
BSchilperoort: From issue #52 it turns out this is already supported by sklearn, and thus this function can be removed.

s2spy/traintest.py
BSchilperoort: I'll clean this class up and implement your suggestions. However, as in our data the anchor_year would always be the group for our use case, let's make this the default?

Peter9192 (approved):
Nice, let's merge! One side note still: we haven't set up mypy yet, right (see #14)? I would be interested to see if mypy confirms this is the right way to use the type hint.

BSchilperoort:
> I would be interested to see if mypy confirms this is the right way to use the type hint.


I quickly installed mypy to test this, and it seems like it does not work, possibly because of sklearn not having type hinting:
```
s2spy\traintest.py:10: error: Skipping analyzing "sklearn.model_selection._split": module is installed, but missing library stubs or py.typed marker
s2spy\traintest.py:17: error: Variable "s2spy.traintest.SplitterClass" is not valid as a type
```

Peter9192:
Let's postpone that to a new PR then

Issue #56, opened 2022-07-27
Supporting multiple target intervals in `map_correlation`.


BSchilperoort:
Currently `rgdr`'s `map_correlation` implementation will calculate the correlation coefficient and p-value between precursor data and target data. It will calculate this over the `anchor_year` dim,  for every dimension in the input data (e.g. `latitude`, `longitude`, `i_interval`). 

However, it does not take into account the Calendar's `n_targets`, more specifically the case where multiple intervals are labelled as target. 
A wrapper function is required to handle this. In the case of multiple target intervals, the precursor data and field data for those intervals are flattened over the `i_interval` dimension. 

The following image serves as a visualization for this concept:
[![image](https://user-images.githubusercontent.com/12114825/181215172-86c33687-51a9-479f-9a54-aa5d556d0bc4.png)](https://nlesc.sharepoint.com/:p:/s/team-beta/EZHgy9Agmd5EieXKWuAUhLgB2eCOxRZaIRfDuStLq8b23w?e=53nCDX)

Peter9192:
By multiple target periods you mean "a target period consisting of multiple data points/intervals", right?

BSchilperoort:
> a target period consisting of multiple data points/intervals

Yes, exactly. I'll clarify this in the issue

PR #57, opened 2022-07-27
Implemented dbscan for RGDR

### LINKED ISSUE #55: missing docstring info in `_map_analysis.py/correlation`
`corr_dim` is not listed in the arguments.
The return type is missing from the type hints.

BSchilperoort:
Implements dbscan clustering functionality. The code included in here is sufficient to reproduce the [prototype_RGDR notebook](https://github.com/AI4S2S/s2spy/blob/6da3233168bc7b083b084a14f7afe5618b8c82c9/notebooks/prototype_RGDR.ipynb).
See the [wip notebook](https://github.com/AI4S2S/s2spy/blob/7a3b0f09b36715717a16f1fe57735e4af0d56bf9/notebooks/working_notebook.ipynb)

<hr>

From @geek-yang:

- [x] Implement DBSCAN clustering function
- [x] Check if the longitude is between -180 : 180 and convert to this range if not
     **not required, haversine uses radians.**
- [x] Handle input with lags dimension to vectorize DBSCAN clustering and keep labels 
     **--> currently supports n_targets == 1, see** #56
- [x] Handle clustering for +/- correlation separately
- [x]  Mask data ( [lat, lon] pairs) using significance (p_value)

When these methods are implemented, we also need to validate the results:

- [x] Check if it is necessary to have weight by area for DBSCAN calculation (the impact of area is already reflected by the haversine distance)
     **weighing by area is required for reducing clusters to the mean of each cluster. This is implemented**
- [x] Add relevant checks for the results and raise warnings (e.g. check whether the results are noise or not).
- [x] Verify DBSCAN clustering results

Note: For RGDR we will want to support grouping the DBSCAN operations over splits, due to the computational intensity when working with full resolution data, see #58

This PR fixes #55

<hr>

Still to be implemented, but out of scope for this PR:
- Investigate currently hard-coded latitude & longitude dimensions. Do we want to make this more dynamic?
- RGDR wrapper that uses the functionality implemented in this PR
    - We need to discuss what structure this will have, e.g. how to show the cluster maps to the user.
- Link between RGDR and train/test splits.
- #56 
- #58 
- Tracking clusters over different splits, as the clusters can differ between folds.


BSchilperoort:
Note that the Python 3.7 Windows build fails. This is due to an issue with the `netcdf4` package, which has become a dependency to be able to use the example data from the proto.

To silence this issue we can specify the version for netcdf4, which might resolve the error. However, for how long do we want to keep supporting Python 3.7? It will reach its end of life in a couple of months.

Peter9192 (changes requested):
Nice work! Perhaps combine all in one module (rgdr.py) and consider adding a `transform` method.

Peter9192 (approved):
Nice, let's follow it up in the next PR

PR #60, opened 2022-07-28
Refactoring calendars to BaseCalendar class and subclasses

### LINKED ISSUE #33: Unintuitive behavior when passing weekly frequencies to calendar.
When passing a pandas frequency to the AdventCalendar such as '2W', Pandas uses calendar weeks to generate the intervals. This causes the situation where the anchor date is not always in an interval.

Example below:
```python
import s2s.time
calendar = s2s.time.AdventCalendar(anchor_date=(5, 10), freq='20W')
calendar.map_years(2019, 2020)
```
returns;
```
>>> i_interval                          0                         1
>>> anchor_year                                                    
>>> 2020         (2019-12-22, 2020-05-10]  (2019-08-04, 2019-12-22]
>>> 2019         (2018-12-16, 2019-05-05]  (2018-07-29, 2018-12-16]
```

Expected behavior: raise an error when frequencies such as calendar week are passed.

### LINKED ISSUE #41: Intervals stored in Calendar object, user workflow possibly unclear
With PR #23 the intervals are stored in the `AdventCalendar` object, and the `map_to_years` function returns `self` (i.e., the calendar) instead of just intervals.

______________
## Previous functionality 
Previously data resampling data was performed in the following way:
```python
calendar = s2spy.time.AdventCalendar()
resampled_data = calendar.resample(data) # intervals retrieved using map_to_data internally
```
And the `map_years` method allowed users to _preview_ what the intervals would look like for the given years.

## Current functionality
With the new changes, the calendar including the generated intervals are returned when calling `map_years`:
```python
calendar = s2spy.time.AdventCalendar()
calendar = calendar.map_years(2020, 2022)
resampled_data = calendar.resample(data)
```
However, when the user then resamples data, `map_to_data` is called internally, and the **intervals the user generated are never used**. The above code block functions the exact same when the second line is removed, as in the `resample` method `map_to_data` will call `map_years`, ignoring any previously set intervals.
________________
## Expected functionality (from my perspective)
It could be more clear to actually **use** the intervals the user generates for resampling data. This makes sure that intervals are only, and explicitly, generated in a single place.

```python
calendar = s2spy.time.AdventCalendar()
calendar = calendar.map_years(2020, 2022)
resampled_data = calendar.resample(data)
data

>>> '*data resampled to the intervals in anchor years 2020, 2021, 2022*'
```

#### If the user did not generate any intervals:
```python
calendar = s2spy.time.AdventCalendar()
resampled_data = calendar.resample(data)

>>> ValueError: 'No intervals to resample data with. Please first generate intervals'
                'with `map_years` or `map_to_data` before calling resample'
```

#### If the data does not fully cover the generated intervals:
```python
calendar = s2spy.time.AdventCalendar()
calendar = calendar.map_years(2020, 2022)
resampled_data = calendar.resample(data)

>>> Warning: 'Insufficient data to resample in 10 intervals, these will contain NaN values.'
```

#### `map_to_data` can still be used, but then as an alternative to `map_years`:
```python
calendar = s2spy.time.AdventCalendar()
calendar = calendar.map_to_data(data1)
resampled_data1 = calendar.resample(data1)
resampled_data2 = calendar.resample(data2)
```

### LINKED ISSUE #50: Resample as a method vs. as a function.
During today's demo session @sverhoeven asked why `resample` is a calendar method, versus just a function that takes a calendar and data as input. 

After some consideration I do agree that this could be a more clear way to do it, and it would demarcate the functionalities of the calendar a bit better (instead of putting everything in it as a method).

This issue is related to #41.

### Current implementation:
```python
import s2spy.time

calendar = s2spy.time.AdventCalendar()
calendar = calendar.map_years(2010, 2020) # does nothing (unused intervals)
data_resampled = calendar.resample(data)
```

### Alternative implementation:
```python
import s2spy.time

calendar = s2spy.time.AdventCalendar()
calendar = calendar.map_years(2010, 2020)
data_resampled = s2spy.time.resample(calendar, data)
```

BSchilperoort:
This PR refactors the current calendars into a BaseCalendar class, where most methods are implemented, and subclasses like `AdventCalendar` or `MonthlyCalendar`. 

As an example I implemented two extra calendars;
`MonthlyCalendar`: where a user inputs an anchor month, and a freq in the form `1M`
`WeeklyCalendar`: where a user inputs an anchor week (as in, week of year), and a freq in the form `1W`

Additionally, the implementation of resample has been moved to a separate module, as `s2spy.time` had gotten too large. Resample now also works as discussed in #50

<hr>
List of changes:

- Resample has also been refactored to a separate function, with its implementations in a separate module.
- Resample will require a mapped calendar as input along with the data that should be resampled.
- Calendars are based of a BaseCalendar class.
- a MonthlyCalendar and a WeeklyCalendar have been implemented, using calendar months and weeks numbers respectively.
- Intervals are not stored in the calendar object, but generated upon request, either with `.show()` or `.get_intervals()`.
- Input data checks that had to be used both in `s2spy.time` and `s2spy._base_calendar` have been moved to `s2spy.utils`.

<hr>

### Demo of functionality

```python
>>> from s2spy.time import MonthlyCalendar
>>> cal = MonthlyCalendar(anchor='Dec', freq='6M')
>>> cal
MonthlyCalendar(month=12, freq=6M, n_targets=1, max_lag=None)

>>> cal.map_years(2020, 2021)
>>> cal.show()
i_interval             (target) 0                     1
anchor_year                                            
2021         (2021 Jun, 2021 Dec]  (2020 Dec, 2021 Jun]
2020         (2020 Jun, 2020 Dec]  (2019 Dec, 2020 Jun]

>>> cal.get_intervals()
i_interval                          0                         1
anchor_year                                                    
2021         (2021-06-30, 2021-12-31]  (2020-12-31, 2021-06-30]
2020         (2020-06-30, 2020-12-31]  (2019-12-31, 2020-06-30]

>>> from s2spy.time import resample
>>> ds_resampled = resample(cal, ds) # Resampled to the intervals shown above

```

<hr>

Still to do:

- [x] Update all docstrings
- [x] ~~Add check to see if the data can fully cover the calendar, and warn the user if it cannot~~ **Out of scope for current PR see #67 **
- [x] ~~Ensure that if the data is insufficient to cover the calendar, NaNs will be used for the empty intervals~~ **Works for xarray data, not for pandas data. Out of scope for this PR. See #67 **

Also:

- decide on setter/getter methods #69 

<hr>

This PR:
Closes #33 
Closes #41 
Closes #50


geek-yang (approved):
Good job!! Lots of progress! The restructuring and refactoring are really clean. I only have some minor comments regarding the code and documentation (see below). Thanks for the effort! 🏆

s2spy/_base_calendar.py
```diff
@@ -0,0 +1,222 @@
...
+    def _get_nintervals(self) -> int:
+        """Calculates the number of intervals that should be generated by _map year.
+
+        Returns:
+            int: Number of intervals for one anchor year.
+        """
+        periods_per_year = pd.Timedelta("365days") / pd.to_timedelta(self.freq)
+        return (
+            (self.max_lag + self.n_targets) if self.max_lag else int(periods_per_year)
+        )
```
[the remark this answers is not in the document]
Peter9192: Oh wait, I see how you did it now. Shouldn't this then be `max_lag + n_target`? And the default `max_lag` is `periods_per_year - n_targets` ?

s2spy/_base_calendar.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def get_intervals(self) -> pd.DataFrame:
+        """Method to retrieve updated intervals from the Calendar object."""
+        if self._mapping is None:
+            raise ValueError(
+                "Cannot retrieve intervals without map_years or "
+                "map_to_data having configured the calendar."
+            )
+        if self._mapping == "data":
+            self._year_range_from_timestamps()
```
Peter9192, line 195:
```suggestion
            self._set_year_range_from_timestamps()
```
? Or: why not call this method directly from `map_to_data`?

s2spy/_base_calendar.py
```diff
@@ -0,0 +1,222 @@
...
+        """Display the intervals nicely."""
+        return self._label_targets(self.get_intervals())
+
+    def __repr__(self) -> str:
+        """String representation of the Calendar."""
+        props = ", ".join(
+            [f"{k}={v}" for k, v in self.__dict__.items() if not k.startswith("_")]
+        )
+        calendar_name = self.__class__.__name__
+        return f"{calendar_name}({props})"
```
Peter9192, line 217: Much better!

s2spy/utils.py
```diff
@@ -0,0 +1,38 @@
+import pandas as pd
+import xarray as xr
+
+
+PandasData = (pd.Series, pd.DataFrame)
+XArrayData = (xr.DataArray, xr.Dataset)
+
+
+def check_input_data(data) -> None:
```
Peter9192, line 9: If this is a generic `util` module, then perhaps this function name should be more specific, e.g. `check_timeseries` or so? Since there are many things you can check wrt input data. Or, if it's so specifically linked to `time.py`, perhaps it makes sense to create a s2s/time/utils.py specific utility folder (and move time.py and _base_calendar.py also to time/time.py).

s2spy/_base_calendar.py
```diff
@@ -0,0 +1,222 @@
...
+
+        intervals = pd.concat([self._map_year(year) for year in year_range], axis=1).T
+
+        intervals.index.name = "anchor_year"
+        return intervals
+
+    @abstractmethod
+    def show(self) -> pd.DataFrame:
+        """Display the intervals nicely."""
+        return self._label_targets(self.get_intervals())
```
Peter9192, line 209: Does a default implementation make sense for an abstract method?

s2spy/time.py
```diff
@@ -90,430 +103,308 @@ def __init__(
...
+    def show(self) -> pd.DataFrame:
+        """Displays the intervals the Calendar will generate for the current setup.
 
         Returns:
-            Pandas DataFrame filled with Intervals of the calendar's frequency,
-            counting backwards from the calendar's anchor_date.
+            pd.Dataframe: Dataframe containing the calendar intervals, with the target
+                periods labelled.
+        """
+        return self._label_targets(self.get_intervals())
```
Peter9192, line 137: Note: same as base calendar.

s2spy/time.py
```diff
@@ -90,430 +103,308 @@ def __init__(
...
-        if isinstance(input_data, PandasData):
-            input_data['target'] = np.zeros(input_data.index.size, dtype=bool)
-            input_data['target'] = input_data['target'].where(
-               input_data['i_interval'] >= self._n_targets, other=True)
+        return self._label_targets(self.get_intervals()).applymap(
+            self._interval_as_weeknr
+        )
+
+
+def resample(
```
Peter9192, line 333: Wouldn't it make more sense for the resample function to be placed with the rest of the implementation inside _resample.py?

Peter9192 (approved):
So far I've had a look at the code only, not tested/played with it yet. In general, I think it looks very good! I have a few comments on the naming and organization of the code. In general, I think it would make sense to create a separate folder for time or calendar so that you get a DRS like this:

```
- s2spy/
  - calendar/
    - _base_calendar.py
    - _weekly_calendar.py etc.  --> alternatively, combine all calendars in _calendar.py
    - _utils.py  --> calendar utils
    - _resample.py --> resampling functionality designed to work specifically with our calendars.

For me, it's fine if you merge this first and then address it in follow-up PRs, for this has already grown quite big.

File: s2spy/_base_calendar.py [hunk 8]
```diff
@@ -0,0 +1,222 @@
...
+        nintervals = self._get_nintervals()
+        periods_per_year = pd.Timedelta("365days") / pd.to_timedelta(self.freq)
+
+        return (
+            (np.ceil(nintervals / periods_per_year).astype(int) - 1)
+            if self.max_lag
+            else 0
+        )
+
+    def map_years(self, start: int = 1979, end: int = 2020):
```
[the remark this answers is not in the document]
BSchilperoort: I do like the consistency between map_years and map_to_data. We can rename this method, but what would be call `map_to_data` then? `set_years_by_data`?

s2spy/_base_calendar.py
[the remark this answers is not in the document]
BSchilperoort: By default the max_lag will be indeed periods_per_year - n_targets. However, max_lag is not a derived parameter like n_intervals is. We could change this function, but that would depend on if we disallow setters (#69)

(I assume you noticed that skipping years is handled by `_get_skip_nyears`.)

s2spy/_base_calendar.py
BSchilperoort: We could if we decide to disallow setting variables after initialization.

Currently the following code works without errors;
```py
>>> calendar = AdventCalendar(freq='30d', max_lag=15)
>>> calendar = calendar.map_to_data(input_data)
>>> calendar.show()
# User sees the calendar and decides on changing the max_lag, to avoid skipping years
>>> calendar.max_lag = 10
>>> calendar.show()
# Year ranges are re-generated from the min, max timestaps, thus the calendar will show correctly.
```
```

File: s2spy/_base_calendar.py (hunk 5 above)
Comment (BSchilperoort) [reply s2spy/_base_calendar.py:209]: I guess not no. I'll change this to not be an abstract method.

File: s2spy/time.py (hunk 6 above)
Comment (BSchilperoort) [reply s2spy/time.py:137]: 👍 removed!

File: s2spy/time.py [hunk 9]
```diff
@@ -90,430 +103,308 @@ def __init__(
...
-    def _resample_pandas(
-        self, input_data: Union[pd.Series, pd.DataFrame]
-    ) -> pd.DataFrame:
-        """Internal function to handle resampling of Pandas data.
+        return pd.Timestamp(year, self.month, 1) + pd.tseries.offsets.MonthEnd(0)
 
-        Args:
-            input_data (pd.Series or pd.DataFrame): Data provided by the user to the
-                `resample` function
+    def _get_nintervals(self) -> int:
```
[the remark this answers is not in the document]
BSchilperoort: The main reason for this is that pandas does not like using monthly frequencies, so a two methods had to be overwritten for the MonthlyCalendar:

- _get_nintervals
- _get_skip_nyears

Removing the default implementation for these would mean that the AdventCalendar and WeeklyCalendar both would have to implement the same methods.

s2spy/utils.py
BSchilperoort: For now I have renamed the function. I'll add refactoring the module/folder structure as an item in an issue, as I do not want to mess with the current PR too much.

s2spy/_base_calendar.py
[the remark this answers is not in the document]
Peter9192: I also wondered about that 😋 Perhaps `infer_years`? Or let `set_years` accept either a tuple of ints or a dataset, and let it infer from the input type what it needs to do? So far I think none of the alternatives are indeed better than keeping the current `map`

s2spy/_base_calendar.py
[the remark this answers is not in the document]
Peter9192: Let's re-evaluate this later. I see how it works now, but somehow it feels a bit cryptic.

s2spy/_base_calendar.py
Peter9192: Okay, let's follow-up in #69

s2spy/time.py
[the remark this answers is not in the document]
Peter9192: Perhaps add that same note as an inline comment in the code where you overwrite them?

BSchilperoort:
Hm. An alternative could also to have the users interface with a single calendar object, e.g. AdventCalendar (or CountdownCalendar or any other suitable name).
Where the following would all be valid inputs;
```py
daycalendar = AdventCalendar(anchor=(31, 12), freq='4d')
weekcalendar = AdventCalendar(anchor=40, freq='2W')
monthcalendar = AdventCalendar(anchor='Jan', freq='1M')
```

PR #65, opened 2022-08-03
Drop python 3.7 and add 3.10

### LINKED ISSUE #64: Drop support for 3.7
Note that the Python 3.7 Windows build fails. This is due to an issue with the `netcdf4` package, which has become a dependency to be able to use the example data from the proto.

To silence this issue we can specify the version for netcdf4, which might resolve the error. However, for how long do we want to keep supporting Python 3.7? It will reach its end of life in a couple of months.

_Originally posted by @BSchilperoort in https://github.com/AI4S2S/s2spy/issues/57#issuecomment-1202804482_

Peter9192:
closes #64

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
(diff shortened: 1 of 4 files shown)


BSchilperoort (approved):
I found no location where the version numbers were not updated. Looks good 👍

PR #66, opened 2022-08-03
Add mypy

Peter9192:
Added to Prospector or run as standalone:

```
# in root dir
prospector

# or
mypy s2spy --ignore-missing-imports
```

This PR also updates the python version for GH actions to 3.10, for there have been substantial improvements in the type hint system recently.

Note: some broken types will be superseded by #60

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,10 +45,10 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v3
         with:
-          python-version: 3.9
+          python-version: 3.10
       - name: Python info
         shell: bash -l {0}
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,8 @@ jobs:
         run: |
           which python3
           python3 --version
+      - name: Run mypy in isolation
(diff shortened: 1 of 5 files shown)


Peter9192:
Getting different results on GH actions and locally... 😕

Peter9192:
> Getting different results on GH actions and locally... 😕

It appears this was due to the pip instruction in the GH Action workflow actually creating a build directory, so there was a duplicate between src and build directories. Excluding build in the prospector config file solved the issue. Now the remaining errors are the same I get locally.

BSchilperoort (approved):
Nice work! I like that mypy is integrated in the prospector configuration, that saves us from locally running another command separately.

I tested mypy on the branch of PR #60, it seems to work without giving errors now.

Edit: mypy did not raise any errors locally. However, it seems that it does on GH actions... 👎 
```
s2spy/_resample.py
  Line: 143
    mypy: error / Argument 1 to "rename" of "Dataset" has incompatible type "str"; expected "Optional[Mapping[Any, Hashable]]" (col 52)

s2spy/rgdr/_rgdr.py
  Line: 81
    mypy: error / Incompatible return value type (got "ndarray[Any, Any]", expected "float") (col 9)
  Line: 128
    mypy: error / Unsupported right operand type for in ("Hashable") (col 57)
```
On my system both prospector (which uses the same configuration as on GH) gives no errors. 

Running `mypy .\s2spy\ --ignore-missing-imports`
returns `Success: no issues found in 9 source files`

PR #74, opened 2022-08-10
Implement train/test splits iterator

### LINKED ISSUE #71: Add iterator feature to traintest
It is necessary to have a iterator in the `traintest.py`, which enables the user to loop through the splits and perform dimensionality reduction (or machine learning) for each train/test group.

geek-yang:
In this PR we implement an iterator for train/test splits, which enable the user to easily fetch train/test data from each split. The potential usecases are listed in issue #71.

- [x] Create generator function for train/test iteration in `traintest.py`
- [x] ~Support `xr.Dataset` and `pd.DataFrame`~ Only xarray
- [x] Add examples in notebooks.
- [x] Support multiple inputs

This PR closes #71. And #75


s2spy/traintest.py
```diff
@@ -8,71 +8,173 @@
...
+                for which calendar.map_years() or calendar.map_to_data(data) has been executed. 
+            key (str): Key by which to group the data. Defaults to 'anchor_year'
+
+        Returns:
+            Initialized traintest_splits
+        """
+        
+        self.splitter = splitter
+        df = resample_bins_constructor(calendar.get_intervals())
+        self.traintest_df = _split_dataframe(splitter, df, key)
```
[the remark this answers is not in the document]
Peter9192: In fact, I don't think we should be storing this at all. To align with sklearn: 

> Randomized CV splitters may return different results for each call of split. You can make the results identical by setting random_state to an integer.

As long as we allow passing multiple arrays into `split_iterate` (for lack of a better name), we can still use the same splits across all inputs.

s2spy/traintest.py
```diff
@@ -8,71 +8,173 @@
...
+                if isinstance(y, xr.Dataset):
+                    train_y = y.sel(anchor_year = self.traintest_xr.sel(split=i) == "train")
+                    test_y = y.sel(anchor_year = self.traintest_xr.sel(split=i) == "test")
+                elif isinstance(y, pd.DataFrame):
+                    train_y = y.loc[self.traintest_df[f'{split}'] == "train"]
+                    test_y = y.loc[self.traintest_df[f'{split}'] == "test"]
+                else:
+                    raise ValueError("Input y should be of type xr.Dataset or pd.DataFrame")
+
+            yield train_Xs, train_y, test_Xs, test_y
```
Peter9192, line 113: If I read this code correctly, this will iterate over the input data args, whereas I would expect a loop over the splits.

Peter9192 (changes requested):

Peter9192:
Finally, green CI! @geek-yang @semvijverberg please briefly review as we've already discussed it yesterday. Since then I've added tests, tests, tests, type annotations, and some additional sanity checks.

geek-yang:
@Peter9192 Nice work! The implementation is very elegant! I have a few very small comments to add. Since this PR is opened by me, I can not approve it. I would prefer @semvijverberg to take care of it. Well done!

s2spy/traintest.py
```diff
@@ -1,105 +1,115 @@
...
+
+        if y is not None and not np.array_equal(y[dim], x[dim]):
+            raise CoordinateMismatch(
+                f"Input arrays are not equal along {dim} dimension."
+            )
+
+        if x[dim].size <=1:
+            raise ValueError(
+                f"Cannot split: need at least 2 values along dimension {dim}"
+            )
```
geek-yang, line 99: Not so sure if this is needed. The splitters in sklearn will take care of this. For instance, if we only have one value and call k-folds method (e.g. n_splits=3) to split it, it will complain:

```py
# create data
n = 8
time_index = pd.date_range("20151020", periods=n, freq="60d")
time_coord = {"time": time_index}
x1 = xr.DataArray(np.random.randn(n), coords=time_coord, name="precursor1")
# resample
calendar = s2spy.time.AdventCalendar(anchor=(10, 15), freq="180d")
calendar.map_to_data(x1) 
x1 = s2spy.time.resample(calendar, x1)
# cross validate
kfold = KFold(n_splits=3)
cv = s2spy.traintest.TrainTestSplit(kfold)
for x1_train, x1_test in cv.split(x1):
    print("Train:", x1_train.anchor_year.values)
    print("Test:", x1_test.anchor_year.values)

>>> ValueError: Cannot have number of splits n_splits=3 greater than the number of samples: n_samples=1.
```

I would suggest to remove this and let the splitters take care of it, as the error message is more informative with different splitter.

s2spy/traintest.py
```diff
@@ -1,105 +1,115 @@
...
+        arrays = iter(arrays)
+        first = next(arrays)
+        return all(np.array_equal(first, rest) for rest in arrays)
+    except StopIteration:
+        return True
 
-    return data
 
+class TrainTestSplit():
+    """Splitters (multiple) xr.DataArrays across a given dimension.
```
geek-yang, line 40:
```suggestion
    """Split (multiple) xr.DataArrays across a given dimension.
```

s2spy/traintest.py
```diff
@@ -1,105 +1,115 @@
...
 
-def split_groups(
-    splitter: Type[BaseCrossValidator],
-    data: Union[xr.Dataset, pd.DataFrame],
-    key: Optional[str] = "anchor_year",
-):
-    """Splits calendar resampled data into train/test groups, based on the input key.
-    As splitter, a Splitter Class such as sklearn's KFold can be passed.
+# Mypy type aliases
+X = Union[xr.DataArray, List[xr.DataArray]]
```
geek-yang, line 17: An interesting finding, `pylint` complains on my machine about single letter variable:
`Variable name "X" doesn't conform to snake_case naming stylepylint(invalid-name)`
But since it follows the vocabulary of machine learning and it doesn't cause confusion, I think that's why in sklearn people also does this. 

But just to be friendly to the users (especially from climate science community) who is new to the ML field, maybe we can add one line in the docstrings at the top: <br>
"In this module, X(or x) refers to training data and y refers to testing data. This is the convention widely used by the machine learning community and for more information please check the machine learning vocabulary of scikit-learn in their code examples: https://scikit-learn.org/stable/tutorial/basic/tutorial.html"

Feel free to add it if you think it is nice to mention.

s2spy/traintest.py
```diff
@@ -8,71 +8,173 @@
...
+# def split_groups(
+#     splitter: Type[BaseCrossValidator],
+#     data: Union[xr.Dataset, pd.DataFrame],
+#     key: Optional[str] = "anchor_year",
+# ):
+#     """Splits calendar resampled data into train/test groups, based on the input key.
+#     As splitter, a Splitter Class such as sklearn's KFold can be passed.
+
+#     For an overview of the sklearn Splitter Classes see:
+#     https://scikit-learn.org/stable/modules/classes.html#module-sklearn.model_selection
```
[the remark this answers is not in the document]
semvijverberg: 👍

s2spy/traintest.py
Peter9192: These Pylint warnings do not show up in prospector because [strictness](https://prospector.landscape.io/en/master/profiles.html#strictness-1) is set to medium. Setting it to high will trigger them in prospector as well.

After our discussion (we agreed to allow certain short/capitalized) variable names, I looked into options for configuring pylint. Interestingly, if I add configuration options in pyproject.toml, they are picked up by pylint run through prospector, but not by pylint run as standalone. Vice versa, adding them to setup.cfg yields the opposite result. By adding .pylintrc, both tools picked up the right configuration. 

@geek-yang I think vscode should also pick up the new rule. Can you verify?

s2spy/traintest.py
Peter9192: In the tests (`test_kfold_too_short`), this produces a much more obscure error message: 
```
--> 325             raise TypeError(
    326                 "Singleton array %r cannot be considered a valid collection." % x
    327             )
```

Hence I think it is worth adding our own, custom error message. However, since sklearn apparently raises different types of errors (ValueError or TypeError), rather than catching it and re-raising, we might be better of with just our own small sanity check. I adapted the message slightly.

Peter9192 (approved):
Since Yang can't approve, I'll do it for him ;-)

s2spy/traintest.py
geek-yang: Thanks for the info. Lessons learned 😉 . Just checked vscode and yes, it picks up the new rule. Thanks for exploring this.

semvijverberg (commented):
Dear @geek-yang @Peter9192,

I'm sorry, I don't have time to properly review this PR. But I'm confident you did a good job ;). 

Cheers,
Sem

PR #77, opened 2022-08-12
Calendar backlog items

### LINKED ISSUE #44: Enable the overlapping of intervals with `max_lag`
Currently, the implementation of `max_lag` doesn't allow the overlapping of intervals for the anchor years, which avoids the information leakage. For instance, if the user define a `advent calendar` in this way

```py
calendar = s2spy.time.AdventCalendar(anchor_date=(11, 30), freq='180d', max_lag = 3)
calendar
```

Then 2019 will be skipped to avoid information leakage.

```
i_interval                          0                         1                          2                         3  
anchor_year                                                       
2020         (2020-06-03, 2020-11-30]  (2019-12-06, 2020-06-03]   (2019-06-09, 2019-12-06]  (2018-12-11, 2019-06-09]  
2018         (2018-06-03, 2018-11-30]  (2017-12-05, 2018-06-03]   (2017-06-08, 2017-12-05]  (2016-12-10, 2017-06-08]  
```

However, for some usecases, the user might allow the overlapping to happen. 

For instance,

```
i_interval                          0                         1                          2                         3  
anchor_year                                                       
2020         (2020-06-03, 2020-11-30]  (2019-12-06, 2020-06-03]   (2019-06-09, 2019-12-06]  (2018-12-11, 2019-06-09]
2019         (2019-06-03, 2019-11-30]  (2018-12-05, 2019-06-03]   (2018-06-08, 2018-12-05]  (2017-12-10, 2018-06-08]  
2018         (2018-06-03, 2018-11-30]  (2017-12-05, 2018-06-03]   (2017-06-08, 2017-12-05]  (2016-12-10, 2017-06-08]  
```

We should provide an option to the user, to allow them have these overlapped intervals.

### LINKED ISSUE #67: Dealing with mismatch between calendar map and data time extent
PR #60 changes resample so that it takes a mapped calendar and resamples the input data based on those intervals.

However, this will require some extra checks and data handling to prevent user error;

- [ ] Add check to see if the data can fully cover the calendar, and warn the user if it cannot.
- [ ] Ensure that if Pandas data is insufficient to cover the calendar, NaNs will be used for the empty intervals
- [ ] Add tests to ensure that both the warning & the NaN filling goes correctly

The second task already works for the xarray implementation, but not yet for Pandas data.

Related to #26

BSchilperoort:
This PR aims to close a few backlog issues about the Calendar/resampling functionality:

- #67 
- #44 
- #26 

List of changes:

- Added check to the input data frequency. Warns the user if the input data frequency is lower than the calendar freq.
- Added a check to see if intervals are missing after resampling (e.g. the data was for 2000 &rarr; 2008, while the calendar was mapped 2000 to &rarr; 2010). Warns the user if there are missing values (NaNs).
- Input data checks are placed in `./utils.py` to avoid cluttering `_resample.py`.
- Added `allow_overlap` kwarg to allow overlapping intervals between anchor_years (and thus not skip years).

Closes #67
Closes #44


BSchilperoort:
@Peter9192 , @geek-yang 
I have currently implemented `allow_overlap` as a kwarg for the calendars. However, this starts to make the amount of arguments for a calendar quite extensive (pylint complains, although it already counts 'self' as an argument in the max number).

An alternative way I can implement this would be by implementing a setter, e.g.:
```py
calendar = AdventCalendar(anchor=(12, 31), freq='1d', max_lag=400)
calendar = calendar.set_allow_overlap(True)
```

geek-yang:
> @Peter9192 , @geek-yang I have currently implemented `allow_overlap` as a kwarg for the calendars. However, this starts to make the amount of arguments for a calendar quite extensive (pylint complains, although it already counts 'self' as an argument in the max number).
> 
> An alternative way I can implement this would be by implementing a setter, e.g.:
> 
> ```python
> calendar = AdventCalendar(anchor=(12, 31), freq='1d', max_lag=400)
> calendar = calendar.set_allow_overlap(True)
> ```

To me, it makes more sense to have this `allow_overlap` as a kwarg, since it is one of the choices that the user needs to make to have the desired calendar. The setter thingy sounds more logical for an added feature. I would prefer to keep it the way it is implemented now.

geek-yang (approved):
Good job! One small thing to be updated. Currently in the notebook `tutorial_time.ipynb`, the example only shows a calendar without overlapping. It is also important to show the user that they can generate a calendar without limit per anchor year. So it is nice to add one more example with `allow_overlap` enabled inside the same notebook

s2spy/_base_calendar.py
```diff
@@ -156,10 +164,10 @@ def _set_year_range_from_timestamps(self):
 
         # ensure that the input data could always cover the advent calendar
         # last date check
-        if self._map_year(map_last_year).iloc[0].right > self._last_timestamp:
+        while self._map_year(map_last_year).iloc[0].right > self._last_timestamp:
             map_last_year -= 1
         # first date check
-        if self._map_year(map_first_year).iloc[-1].left < self._first_timestamp:
+        while self._map_year(map_first_year).iloc[-1].right <= self._first_timestamp:
```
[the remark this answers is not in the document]
BSchilperoort: The `while` loop is required, as with `max_lag` the calendar can extend further backwards in time.

Although that only applies to `map_first_year`, and not `map_last_year`, so I can change line 167.

s2spy/_base_calendar.py
```diff
@@ -156,10 +164,10 @@ def _set_year_range_from_timestamps(self):
 
         # ensure that the input data could always cover the advent calendar
         # last date check
-        if self._map_year(map_last_year).iloc[0].right > self._last_timestamp:
+        while self._map_year(map_last_year).iloc[0].right > self._last_timestamp:
```
Peter9192, line 167:
```suggestion
        while self._map_year(last_year).iloc[0].right > self._last_timestamp:
```
Would it make sense to rename this? I got a bit confused by the "double mapping".

s2spy/_base_calendar.py
```diff
@@ -196,8 +204,9 @@ def get_intervals(self) -> pd.DataFrame:
         if self._mapping == "data":
             self._set_year_range_from_timestamps()
 
+        skip_years = 0 if self.allow_overlap else self._get_skip_nyears()
```
Peter9192, line 207: Perhaps move `skip_years = 0 if self.allow_overlap` into `self._get_skip_nyears()` to keep all the logic together?

s2spy/time.py
```diff
@@ -70,12 +70,13 @@
 class AdventCalendar(BaseCalendar):
     """Countdown time to anticipated anchor date or period of interest."""
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
```
Peter9192, line 73: I think I'd prefer a setter.

Peter9192 (approved):
Nice work. I don't think it closes #26, though. That issue, as I understand it, also suggests that we actually deal with the resulting NaN values, instead of only warning the user.

s2spy/_base_calendar.py
BSchilperoort: Looking at it, `min_year` and `max_year` might be a more appropriate name. I find `first` and `last` a bit ambiguous with the BackwardCalendar.

s2spy/time.py
BSchilperoort: Hm. How about keeping `max_lag` and `allow_overlap` together then?
For example:
```py
calendar = MonthlyCalendar(anchor='Dec', freq='1M')
calendar.set_max_lag(20, allow_overlap=True)
```

Peter9192: I like that very much

geek-yang: Also sounds good to me, to put them together.

PR #83, opened 2022-08-17
Corrections to the RGDR methods

### LINKED ISSUE #80: Modify interface of RDGR to conform better to sklearn
Currently RGDR works like this:

```py
target_timeseries = ...
precursor_field = ...
rgdr = s2spy.RGDR(target_timeseries)
clustered_data = rgdr.fit(precursor_field)  # does what sklearn fit_transform does
```

It might be nice to remodel it such that it would be used like this:

```py
target_timeseries = ...  # y
precursor_field = ...  # X
X_train,  X_test, y_train, y_test = s2spy.train_test_split(precursor_field, target_timeseries)
rgdr = s2spy.RGDR()
rgdr.fit(X_train, y_train)

# or e.g.
pipe = Pipeline([('rgdr', RGDR()), ('svc', SVC())])
pipe.fit(X_train, y_train)
pipe.score(X_test, y_test)
```

### LINKED ISSUE #82: Return correlation / p_values from RGDR
Currently the correlation and p_values are not accessible by the user. The user can only visualize the those values via RGDR built-in plot functions. The users may need to investigate/store these values in case they want to check the correlation of the precursor fields and target timeseries for all the `i_intervals`. So RGDR should return correlation and p_values.

BSchilperoort:
This PR:

- Corrects the functionality of `.fit`
- Moves the `timeseries` kwarg into `.fit` instead of `__init__`
- Adds the `get_correlation` and `get_clusters` methods to be able to access the intermediate steps of RGDR.

Closes #82
Closes #80


Peter9192 (changes requested):
Nice & quick. I merged it into our today's script. Would be nice to have a fit_tranform method as well, see https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA.fit for reference. Otherwise we get:

```
rgdr.fit(
    precursor_train.isel(i_interval=lag), 
    target_train
).transform(precursor_train.isel(i_interval=lag))
```

Peter9192 (approved):
Thanks! Used it in #84

Issue #91, opened 2022-08-24
plot_correlation() requires recalculation of correlation map


semvijverberg:
Issue: plot_correlation() requires recalculation of correlation map, which hampers interpretability that might be wanted _after_ fitting a complex pipeline. The RGDR method can be used a 'brute' forcing dimensionality reduction approach, in which, a feature selection method or explainable AI method might identify some important features. It would then be nice if the user is able to visualize - in hindsight - which are these important features on the map. What I would like to be able to do is:

```
import xarray as xr
import matplotlib.pyplot as plt
import s2spy.time
import s2spy.rgdr
from s2spy import RGDR

file_path = '../tests/test_rgdr/test_data'
field = xr.open_dataset(f'{file_path}/sst_daily_1979-2018_5deg_Pacific_175_240E_25_50N.nc')
target = xr.open_dataset(f'{file_path}/tf5_nc5_dendo_80d77.nc')

cal = s2spy.time.AdventCalendar((8, 31), freq = "30d")
cal = cal.map_to_data(field)
field_resampled = s2spy.time.resample(cal, field)
target_resampled = s2spy.time.resample(cal, target)

rgdr = RGDR()
rgdr.fit(precursor_field, target_timeseries)

# fit model
# do features selection or feature importance analysis
# Desire to investigate important feature identified by features selection or feature importance analysis with label "lag:1_cluster:-2"

rgdr.plot_correlation(lag=1)
rgdr.plot_labels(lag=1, label=-2)
```

geek-yang:
The related PR is merged and this issue can be closed now.

PR #94, opened 2022-08-25
Now storing corr, pval & cluster_map into rgdr instance

semvijverberg:
Concerning issue #91 .

To allow users to inspect the output of RGDR in hindsight, I have added corr_map, pval_map, and cluster_map to the instance that is created after fitting, i.e., RGDR().fit(field, timeseries).

This means that the hidden rgdr._clusters which was already stored on the instance is now visible as 'rgdr.cluster_map'. 

In the tutorial_RGDR.ipynb, I have a added a single cell, where corr_map is plotted after calling rgdr.fit().


s2spy/rgdr/rgdr.py
```diff
@@ -508,8 +508,9 @@ def fit(self, precursor: xr.DataArray, timeseries: xr.DataArray):
         masked_data = masked_spherical_dbscan(
             precursor, corr, p_val, self._dbscan_params
         )
-
-        self._clusters = masked_data.cluster_labels
+        self.corr_map = corr
+        self.pval_map = corr
+        self.cluster_map = masked_data.cluster_labels
```
Peter9192, line 513:
```suggestion
        self.corr_map = corr
        self.pval_map = p_val
        self.cluster_map = masked_data.cluster_labels
```

s2spy/rgdr/rgdr.py
```diff
@@ -343,7 +343,7 @@ def __init__(
             min_area_km2 (float): The minimum area of a cluster. Clusters smaller than
                 this minimum area will be discarded.
         """
-        self._clusters = None
+        self.cluster_map = None
```
Peter9192, line 346: also initialize other new attrs as None.

s2spy/rgdr/rgdr.py
```diff
@@ -508,8 +508,9 @@ def fit(self, precursor: xr.DataArray, timeseries: xr.DataArray):
         masked_data = masked_spherical_dbscan(
             precursor, corr, p_val, self._dbscan_params
         )
-
-        self._clusters = masked_data.cluster_labels
+        self.corr_map = corr
+        self.pval_map = corr
```
Peter9192, line 512: New public API attributes should be documented.

Peter9192 (dismissed):
Please see comments on code.

notebooks/tutorial_RGDR.ipynb
```diff
@@ -70,7 +70,7 @@
    "outputs": [
```
geek-yang, line 70: <p>To be complete, also add <code>rgdr.pval_map[0].plot()</code>, preferably in the same cell parallel to the correlation map.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/AI4S2S/s2spy/pull/94/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

s2spy/rgdr/rgdr.py
geek-yang: A small note @semvijverberg , if you have linters configured in your coding environment, then it will warn you that "Attribute 'corr_map' defined outside __init__pylint(attribute-defined-outside-init)". So it is nice to work with an editor with python linters/extensions configured 😜 .

s2spy/rgdr/rgdr.py
semvijverberg: Thanks for the comments Peter! 

How do I update the API documentation? I was kind of assuming that this would go automatically.

s2spy/rgdr/rgdr.py
semvijverberg: Done.

s2spy/rgdr/rgdr.py
semvijverberg: Oeps! Thanks for spotting.

s2spy/rgdr/rgdr.py
Peter9192: See e.g. https://stackoverflow.com/a/3051356

notebooks/tutorial_RGDR.ipynb
geek-yang: New cell added in the notebook to show the `rgdr.pval_map[0].plot()` now.

semvijverberg: <p>As you can see, the xr.DataArray cluster_map first needs to be transformed to integers for plotting.</p><p> </p><p>The function cluster_labels_to_ints copies the values (dtype: strings) and converts the values to dtype: int. This converted array is added as a coordinate (the original cluster_map.values remain strings).</p><p> </p><p>Maybe it would be better if the cluser_map.values are integers by default (making the plotting intuitive) and that the cluster_labels are added as a coordinate.</p><p> </p><p>These cluster_map coordinates can then be used to do the groupby(labels).mean() (https://github.com/AI4S2S/s2spy/blob/8ff64ba23880bab9e24adf1aa268c328e1f79fc6/s2spy/rgdr/rgdr.py#L517-L522). The latter function will need to be slightly adapted to support this.</p><p> </p><p>What do you think?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/AI4S2S/s2spy/pull/94/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

s2spy/rgdr/rgdr.py
geek-yang: The attributes are documented in the docstrings now.

notebooks/tutorial_RGDR.ipynb
geek-yang: <p>I see the point. I agree that in this case integers are better than unicode strings. But I think it is better to look into it and check the code when these labels are generated rather than just add a conversion in the end. So I prefer that we raise an issue for it and deal with it in another PR. Let's merge this PR for now.</p>

geek-yang (approved):
@semvijverberg Thanks for the changes. Looks good to me for now. Changes are approved.

PR #96, opened 2022-09-01
Prepare minor release 0.2.0

geek-yang:
- Update changelog
- Update setup.cfg
- Bump version to 0.2.0

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -33,7 +33,7 @@ authors:
     affiliation: "Vrije Universiteit Amsterdam"
 
 date-released: 2022-09-01
-version: "0.2.0""
+version: "0.2.0"
 repository-code: "https://github.com/AI4S2S/s2spy"
 keywords:
   - s2s
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -33,7 +33,7 @@ authors:
     affiliation: "Vrije Universiteit Amsterdam"
 
 date-released: 2022-09-01
-version: "version: "0.2.0""
+version: "0.2.0""
 repository-code: "https://github.com/AI4S2S/s2spy"
 keywords:
   - s2s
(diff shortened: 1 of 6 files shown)


CITATION.cff
```diff
@@ -32,9 +32,8 @@ authors:
     given-names: Jannes
     affiliation: "Vrije Universiteit Amsterdam"
 
-date-released: 20??-MM-DD
-doi: <insert your DOI here>
-version: "0.1.0"
+date-released: 2022-09-01
+version: "version: "0.2.0""
```
Peter9192, line 36: There's a double "version" here, is that correct?

geek-yang: Thanks for spotting this. Yep, this is a bug caused by `bump2version`. When I run it, it added these double versions thingy to some files...Had a discussion with Sarah but we don't know why...So I simply delete these double versions. But forgot this one. I will investigate this a bit.

geek-yang: Quite strange...`bump2version` just not work...When I simply run `bump2version 0.2.0`, it keeps causing double versions and it only gives "0.1.0". It seems `bump2version` fails to recognize the version, although `.bumpversion.cfg` already exists. Also did not find anything useful after some googling. I will ask Stefen tomorrow. For now, I can only change it manually 😢 .

CITATION.cff
```diff
@@ -32,9 +32,8 @@ authors:
     given-names: Jannes
     affiliation: "Vrije Universiteit Amsterdam"
 
-date-released: 20??-MM-DD
-doi: <insert your DOI here>
-version: "0.1.0"
+date-released: 2022-09-01
+version: "0.2.0""
```
SarahAlidoost, line 36:
```suggestion
version: "0.2.0"
```

SarahAlidoost (approved):
@geek-yang looks good to me, just a small fix on the `CITATION.cff`. Also, please submit an issue about the usage of  `bump2version`.

geek-yang:
> @geek-yang looks good to me, just a small fix on the `CITATION.cff`. Also, please submit an issue about the usage of `bump2version`.

Thanks for the feedback. This #98 issue relates to the bug of `bump2version`.

PR #97, opened 2022-09-01
Fix image display on ReadtheDoc and pypi description

geek-yang:
This will fix the broken images on readthedocs (also for description on pypi).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<img width="300" alt="Logo" src="./docs/assets/images/ai4s2s_logo.png">
+<img width="300" alt="Logo" src="https://github.com/AI4S2S/s2spy/blob/main/docs/assets/images/ai4s2s_logo.png">
 
 # s2spy
 
@@ -56,7 +56,7 @@ python3 -m pytest
 ## Getting started
 `s2spy` provides end-to-end solutions for machine learning (ML) based S2S forecasting.
 
-![workflow](./docs/assets/images/workflow.png)
+![workflow](https://github.com/AI4S2S/s2spy/blob/main/docs/assets/images/workflow.png)
 
 ### Datetime operations & Data processing
 In a typical ML-based S2S project, the first step is always data processing.  A calendar-based datetime module `time` is implemented for time operations. For instance, a user is looking for predictors for winter climate at seasonal timescales (~180 days). First, a `calendar` object is created using `AdventCalendar`:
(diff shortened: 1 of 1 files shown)


SarahAlidoost:
@geek-yang it seems that ReadTheDocs doesnot display images from github. 

In ReadTheDocs, the `'.'` directory is `_build/html`. So, relative path `./docs/assets/images` in the readme does not work in `index.html`. One workaround is to move `assets/images` to one directory up and rename `assets` to `_static`. Then in `docs/conf.py`, we can add  `html_static_path = ["../_static"]`. This will copy the folder `images` to `_build/html/_static`. Hence the relative path `./_static/images` works for both readme and ReadTheDocs, for more info, see  [html_static_path](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=STATIC_URL#confval-html_static_path), and [html_extra_path](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=STATIC_URL#confval-html_extra_path).

geek-yang:
> @geek-yang it seems that ReadTheDocs doesnot display images from github.
> 
> In ReadTheDocs, the `'.'` directory is `_build/html`. So, relative path `./docs/assets/images` in the readme does not work in `index.html`. One workaround is to move `assets/images` to one directory up and rename `assets` to `_static`. Then in `docs/conf.py`, we can add `html_static_path = ["../_static"]`. This will copy the folder `images` to `_build/html/_static`. Hence the relative path `./_static/images` works for both readme and ReadTheDocs, for more info, see [html_static_path](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=STATIC_URL#confval-html_static_path), and [html_extra_path](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=STATIC_URL#confval-html_extra_path).

@SarahAlidoost Thanks a lot for your suggestion! I think this would work. But then I notice that for the description on PyPI, it can't read package distributions for the images, which means relative links will not work. See this [post](https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github). I also checked ewatercycle and dianna. Both of them use absolute links to images in their readme and create separate .rst file for readthedocs. I think I will stick to their solution. But thank you so much for the solution! Just learn quite a lot from it.

geek-yang:
It is fixed by using the raw external link provided by github.

Issue #79, opened 2022-08-15
Storing resampled data issue due to pd.Interval dtype


semvijverberg:
When I try to store a resample xr.DataArray:

![image](https://user-images.githubusercontent.com/35559713/184626823-c1f032e0-311e-42aa-b86d-3fe2df44ff69.png)

 I get the following error:
 
![image](https://user-images.githubusercontent.com/35559713/184626951-50f26b4a-b2a3-4ac2-8ae0-b2a0218e4ecf.png)

For completeness, this is the whole dataset:
![image](https://user-images.githubusercontent.com/35559713/184627088-030dc522-8446-428d-baf2-ca19c25720c9.png)

geek-yang:
Found some similar issues after a bit of googling https://stackoverflow.com/questions/54848671/using-xarray-groupby-bins-results-in-coordinates-that-are-an-object-and-cant-be. Apparently we are not the only one. There is already an issue in xarray about supporting `interval index` https://github.com/pydata/xarray/issues/2847. 

Currently in our implementation, the interval datatype is object, which is not supported. But from the discussions in those issues, it seems xarray does support certain interval types. Maybe we can try to play a bit with different interval types generated by pandas? (e.g. https://pandas.pydata.org/docs/reference/api/pandas.interval_range.html)

BSchilperoort:
Seeing as we do not really use the intervals ourselves after applying the calendar, we can just decide to not include them in the DataArray. 

If we do want to keep them, we can split the intervals up into two coordinates; `interval_left` and `interval_right` (or `interval_start` and `interval_end`. These will just be timestamps and can therefore be stored in a netCDF.

Peter9192:
Perhaps we could convert them to normal times (either 'left',  'center', or 'right'), and include a `bounds` array of shape `2n`, where `n` is the length of the calendar. This is how it's done in the CF/CMOR standards.

Peter9192:
See https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries and e.g. https://cf-xarray.readthedocs.io/en/latest/bounds.html?highlight=bounds

geek-yang:
[...]

PR #104, opened 2022-09-21
Util to convert intervals to bounds, issue #79

BSchilperoort:
As Sem noticed (#79), having data of the type `pd.Interval` will prevent users from saving their `xr.Dataset` to netCDF. Therefore we decided on converting the intervals to bounds, following CF conventions.

This PR adds a utility function that will convert the intervals of the resampled data to bounds, and this function is called at the end of the `resample_xarray` routine.

To ensure that the calendar-resampled data can always be written to a netcdf file, a test was added for this functionality.

<hr>

The resampled data will now look like the following. Changes are in bold.

<pre>
<xarray.Dataset>
Dimensions:      (anchor_year: 39, i_interval: 12, latitude: 5, longitude: 13,
                  <b>bounds: 2</b>)
Coordinates:
  * anchor_year  (anchor_year) int32 1980 1981 1982 1983 ... 2015 2016 2017 2018
  * i_interval   (i_interval) int64 0 1 2 3 4 5 6 7 8 9 10 11
  * latitude     (latitude) float64 47.5 42.5 37.5 32.5 27.5
  * longitude    (longitude) float64 177.5 182.5 187.5 ... 227.5 232.5 237.5
    index        (anchor_year, i_interval) int64 0 1 2 3 4 ... 464 465 466 467
    interval     (<b>bounds</b>, anchor_year, i_interval) datetime64[ns] 1980-08-01 ...
    target       (i_interval) bool True False False False ... False False False
Data variables:
    sst          (latitude, longitude, anchor_year, i_interval) float64 284.2...
</pre>


geek-yang (approved):
Nice implementation! Works well. Just go ahead and merge it! Thanks for the effort 👍 .

PR #107, opened 2022-09-26
Calendar visualizations using `bokeh`

BSchilperoort:
This PR adds "bokeh" as an alternative backend for the visualization utility.

The advantage of bokeh over matplotlib is that it allows for interactive plots in jupyter notebooks. Tooltips can then provide more detail on each interval.

This PR adds a basic implementation of the visualization. More additions are planned, but would take too much time/attention at the moment (see #109 ).

<hr>

If multiple anchor years are to be displayed, the plot looks like the following:
image

<hr>

With the `relative_dates` kwarg, the calendar can either be aligned to the anchors or show absolute dates of the intervals:
image


BSchilperoort:
Note that the linter fails only because of a known mypy bug (https://github.com/python/mypy/issues/13627)

BSchilperoort:
Working on some refactoring and streamlining of the code, along with adding your (live) suggestions 👍

notebooks/tutorial_time.ipynb
```diff
@@ -272,6 +272,62 @@
     "larger_calendar.visualize(add_freq=True)"
```
geek-yang, line 272: <p>This needs to be fixed. It runs well locally on my machine. I think you haven't updated/run this notebook after changing your code. It must be <code>from .bokeh_plots import bokeh_visualization</code> , which is correct in your code. Just run it again.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/AI4S2S/s2spy/pull/107/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

geek-yang (approved):
Nice work! The implementation is also quite elegant. Only for the file `plot.py`, I think the user doesn't need it and it can also be a hidden thingy, just like `_bokeh_plots.py`.

Changes are approved. Just feel free to merge it to main after addressing the minor comments.  Thanks for the nice work!

notebooks/tutorial_time.ipynb
BSchilperoort: Yeah this seems to have gone wrong. Thanks for noticing!

Issue #100, opened 2022-09-14
Add more flexibility to BaseCalendar by enhancing the basic building block


geek-yang:
All calendars are generated based on the `BaseCalendar`. Currently we use `pd.interval_range` function to initialize the `BaseCalendar`. However, this makes it more difficult to work with gaps and limits its flexibility to support/construct more calendar types.

We can think about a basic building block which allows the developer (not visible to the user) to define gaps, targets and precursors easily, for instance like what is shown in the figure below:

![220914_explain_ai4s2s_time](https://user-images.githubusercontent.com/25911757/190196002-44a516c9-7597-4b34-8988-f0965f7e117e.png)

To define this base calendar, we can add more variables to provide more flexibility (e.g. `target freq`, `number of targets`, `gap time`, `gap freq`, `precursor freq`, `number of precursors`). 

This can help with any type of calendar we would like to build, for instance, an `AdventCalendar` with gap (and allows target and precursor to have different frequency):

![image](https://user-images.githubusercontent.com/25911757/190198483-e9dae8c2-e56e-4016-9581-bbb2e84429d0.png)

Or a `IntervalCalendar` with multiple gaps:

![image](https://user-images.githubusercontent.com/25911757/190197839-36c81d82-b11f-4b2e-ae98-946dd3f39789.png)

We only show the user high level API of calendars (e.g. the API for `AdventCalendar` will remain the same). We can also provide a "guru" version of calendar which gives user full space to define super complex calendar if they want.

But technically, we cannot rely on `pd.interval_range`. But we can still use the datetime module of pandas, to generate intervals and manage time, which can be further discussed.

Peter9192:
Thanks for elaborating on this. Perhaps for constructing multiple calendars out of small building blocks, it might also help to define the target and precursor periods as unique blocks:

```py
precursor = s2spy.PrecursorPeriod(frequency=..., start=...)  # or frequency and end, or start and end
target = s2spy.TargetPeriod(frequency=..., start=....)

ltcalendar = s2spy.Calendar(target, precursor, lead_time='14d')  # or keep using lag instead of lead_time

# Create a compound calendar for searching through a range of lags/lead_times
calendars = []
for lead_time in range(14, 64, 7):
     calendars.append(s2spy.Calendar(target, precursor, lead_time=lead_time))
lagsearch_calendar = s2spy.concat(calendars)  # modelled after pandas/xarray

# Or, alternatively, create a high-level constructor for a compound calendar
[...]

PR #108, opened 2022-09-28
Custom calendar with building blocks

geek-yang:
To enhance the calendar and provide more flexibility, we change the way for building the calendars by adding basic building blocks and concatenating these blocks to a calendar object. The changes will be made in this PR and these include:

- Add basic building block `Period` and create `target_period` and `precursor_period` in `time` module
- Support gap in calendars
- Add a `CustomCalendar` for advanced users to build their own calendar

It closes issue #100 .


s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+                total_length + 1, unit="D"
+            )
+            end_date = self._get_anchor(year) - pd.Timedelta(1, unit="D")
+            # generate date ranges
+            date_range = pd.date_range(start=start_date, end=end_date)
+            date_range = date_range[::-1]
+        # generate intervals based on the building blocks
+        intervals = []
+        # pointer to the index of date for right boundary of interval
+        left_date_index = 0
```
BSchilperoort, line 410: This pointer system storing the days from the anchor date as an `int` does not seem like the most efficient or robust way to do it. Why not base it completely off `pd.Timestamp` objects?

```py
intervals = []
date_pointer = anchor_timestamp # literal pd.timestamp, based on anchor date + year
for precursor in precursor_periods:
    left = date_pointer - pd.Timedelta(precursor.length + precursor.gap)
    intervals.append(pd.Interval(
        left=left,
        right=date_pointer,
        closed='left'))
    date_pointer = left

date_pointer = anchor_timestamp
for target in target_periods:
    # Similar but then build towards the right.
```

This would also make adding/subtracting those single days unnecessary.

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+            if period_block.gap < 0:
+                raise ValueError(
+                    "First appended period block should not contain a negative gap."
+                )
+        else:
+            if (period_block.gap + list_period[-1].length) < 0:
+                raise ValueError(
+                    "Appended period block has negative gap larger than"
+                    + " the length of precedent period block."
+                )
```
BSchilperoort, line 368: I thought we would not have these guardrails in place, to make our lives easier and give people complete freedom in building the calendars.

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+
+        Note that two rules are applied to avoid information leakage and disorder:
+            - period block next to the anchor date should not have negative gap
+            - if a appended period block has negative gap, its absolute value must
+             be smaller than the length of the precedent period block
+        """
+        self.month = anchor[0]
+        self.day = anchor[1]
+        self._list_target = []
+        self._list_precursor = []
```
BSchilperoort, line 318: Why not just name these `self._targets` and `self._precursors`?

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
         return self._label_targets(self.get_intervals()).applymap(
             self._interval_as_weeknr
         )
+
+
+class CustomCalendar(BaseCalendar):
+    """Build a calendar from sratch with basic construction elements."""
+
+    def __init__(self, anchor: Tuple[int, int] = (11, 30)):
```
BSchilperoort, line 304: Let's not have a default value for the anchor. It's very arbitrary and users will always have to set it.

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+            pd.Timestamp: Timestamp at the end of the anchor_years interval 0.
+        """
+        return pd.Timestamp(year, self.month, self.day)
+
+    def append(self, period_block):
+        """Append target/precursor periods to the calendar."""
+        if self._end_indentifier:
+            warnings.warn(
+                "Calerdar already exists and this overwrites the old calendar."
+            )
```
BSchilperoort, line 343: Is this warning necessary? If you change the calendar settings the intervals can easily be regenerated.

After calling map_years or map_to_data the minimum/maximum extents are stored in the calendar already, so the calendar can be generated after that. Even the order `initialize -> map -> append periods -> show()` should be no problem.

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+            self._append_check(self._list_target, period_block)
+            self._list_target.append(period_block)
+            # count length
+            self._total_length_target += period_block.length + period_block.gap
+
+        else:
+            self._append_check(self._list_precursor, period_block)
+            self._list_precursor.append(period_block)
+            # count length
+            self._total_length_precursor += period_block.length + period_block.gap
```
BSchilperoort, line 354: These "total lengths" will have to be refactored in some way when we move away from having days as integers

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+            The calendar mapped to the input start and end year.
+        """
+        if start > end:
+            raise ValueError("The start year cannot be greater than the end year")
+
+        self._first_year = start
+        self._last_year = end
+        self._mapping = "years"
+        return self
+
```
BSchilperoort, line 457: The `map_to_data` method seems to be missing here.

BSchilperoort (changes requested):
Hi Yang, nice work. I did identify some issues that should probably be addressed before merging.

Also, I just had the idea to make `allow_overlap=False` work differently. Why don't we first generate the intervals, and then throw out anchor years if there is overlap. This avoids having to do complex calculations when working with non-day Timedeltas.
For the plotting utilities we already require a `get_overlapping_periods` method on the calendar anyway (to display _which_ intervals overlap).

Some pseudocode:
```py
intervals = self.get_intervals()
overlap, overlapping_intervals = assert_overlap(intervals)
while overlap:
    self.skip_years = 1
    intervals = self.get_intervals()
    overlap, _ = assert_overlap(intervals)
```
Of course this makes generating the calendar a bit slower, but I don't think the calendar generation will ever be a bottleneck. On the upside; it makes our code a lot simpler.

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+
+def target_period(length, gap: int = 0):
+    """Instantiate a build block as target period. """
+    return Period(length=length, gap=gap, target=True)
+
+
+def precursor_period(length, gap: int = 0):
+    """Instantiate a build block as precursor period."""
+    return Period(length=length, gap=gap, target=False)
```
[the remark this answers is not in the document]
BSchilperoort: This basically works as a drop-in replacement by the way, requiring no changes elsewhere ;)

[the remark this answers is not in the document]
geek-yang: Thanks for the comment, Bart. Actually I thought about it at the beginning, but then felt that it might be an overkill in `time.py`...But I think the big advantage of this design is, we can be more flexible in case later on we might want to add something to Precursor Period or Target Period. I think we can put it this way.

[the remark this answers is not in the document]
BSchilperoort: Right now it's not really needed functionally, but it does indeed add flexibility if we need it in the future. Additionally, it's more clear to the user that they are a class. Because the function `target_period()` returns a class `Period` already.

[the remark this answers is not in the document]
geek-yang: That makes sense to me. Good point! 😉

s2spy/time.py
```diff
@@ -304,3 +296,237 @@ def show(self) -> pd.DataFrame:
...
+        left_date_index = 0
+        # loop through all the building blocks to
+        for block in list_periods:
+            left_date_index += block.gap
+            # pointer to the index of date for left boundary of interval
+            right_date_index = left_date_index + block.length - 1
+            if is_target:
+                intervals.append(
+                    pd.Interval(
+                        date_range[left_date_index] - pd.Timedelta(1, unit="D"), # open left close right
```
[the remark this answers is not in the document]
geek-yang: It is not a bug 😅. I have mentioned this in teams in our conversation. Perhaps you missed that comment. And my thought was partly the same as what you put here. We had closed right side before because of the `Adventcalendar`, which has the anchor date placed in the end. Now we have it in the middle. If we stick to the same "left open, right closed" rule, then the anchor date is invisible when you display the calendar. That is how the code works now.

I have considered to change it to "left closed, right open", since we move the anchor date in the middle. Given the conflict of two designs, we need to make a choice. Well, there are several things we should take into account:
- We expect users to use `AdventCalendar` and its variants more often
- This custom calendar will become the base calendar and it won't be shown to the users in most of the cases
- The `CustomCalendar` will only be shown to the advanced users

In other words, `AdventCalendar` has priority. So I decide to keep "left open, right closed" rule. That's why. I think we'd better have a discussion about it.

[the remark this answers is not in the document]
geek-yang: What I have in mind now, is we make this `CustomCalendar` "left closed, right open", and when update `AdventCalendar` based on `CustomCalendar`, we change the results to be "left open, right closed". So we take care of both cases.

s2spy/time.py
geek-yang: Sounds good. I will change these. 😸

s2spy/time.py
geek-yang: Thanks for the suggestion. Corrected!

s2spy/time.py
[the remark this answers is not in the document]
BSchilperoort: > What I have in mind now, is we make this `CustomCalendar` "left closed, right open", and when update `AdventCalendar` based on `CustomCalendar`, we change the results to be "left open, right closed". So we take care of both cases.

Yes, I believe that as we're switching the anchor to be on the left side of the interval, the intervals _have_ to be left closed, right open. Refactoring the AdventCalendar/other calendars to this system can indeed come later.

s2spy/time.py
geek-yang: Yep, as suggested by Peter, not something perfect, but simpler and easy to use as a starting point. So I think we can have `map_to_data` in next PR. Do you agree?

s2spy/time.py
geek-yang: I thought these are really the bottom lines...But I will just, let it go 🙈, haha. Complete freedom unlocked 😜.

s2spy/time.py
geek-yang: Yep, indeed it will not cause trouble to the workflow. Only in case the user accidently calls the appending function after the calendar is already generated. But I agree, it seems more logical that a user calls it on purpose. Let's remove this warning.

s2spy/time.py
geek-yang: Yessss! Thanks @BSchilperoort ❤️! I should have thought about it! Brilliant idea! This really solves the problem for annoying index manipulation. Just implemented this. Now the concatenation looks very clean. 👍

s2spy/time.py
geek-yang: After refactoring the `_concatenate_periods`, for now these "total lengths" are still needed for `_get_skip_nyears`. Later on when we refactor the `allow_overlap` function, we can remove it.

geek-yang:
> Hi Yang, nice work. I did identify some issues that should probably be addressed before merging.

Thanks a lot for the review @BSchilperoort. The comments are super helpful. I made many changes based on your comment. Could you please take a look again?

> Also, I just had the idea to make `allow_overlap=False` work differently. Why don't we first generate the intervals, and then throw out anchor years if there is overlap. This avoids having to do complex calculations when working with non-day Timedeltas. For the plotting utilities we already require a `get_overlapping_periods` method on the calendar anyway (to display _which_ intervals overlap).
> 
> Some pseudocode:
> 
> ```python
> intervals = self.get_intervals()
> overlap, overlapping_intervals = assert_overlap(intervals)
> while overlap:
>     self.skip_years = 1
>     intervals = self.get_intervals()
>     overlap, _ = assert_overlap(intervals)
> ```
> 
> Of course this makes generating the calendar a bit slower, but I don't think the calendar generation will ever be a bottleneck. On the upside; it makes our code a lot simpler.

Cool! Good idea. I think this is really good since then the timedelta of calendar will not bother us. But I would suggest we create a new issue and refactor it in another PR. We will need to write some tests and update notebooks for it. This will be a bit too much in this PR, since our goal in this PR is to have a starting point (and we still need to get back all CD/CI tests for it). So let's do create a new item for it. Thanks for the cool idea 👍.

BSchilperoort:
Hi Yang, just to get a full overview, the next steps are the following, right?

- [ ] Refactor `i_interval` to use the new indexing system we discussed (targets have positive numbers counting up from the anchor, precursors have negative numbers counting down away from the interval)
- [ ] Change the anchor kwarg (#114)
- [ ] Refactor `skip_n_years` (#112)
- [ ] Add tests for the CustomCalendar

After these we can continue with the items we discussed on the Board (replacing the BaseCalendar w/ the CustomCalendar).

s2spy/time.py
```diff
@@ -314,13 +317,11 @@ def __init__(self, anchor: Tuple[int, int] = (11, 30)):
         """
         self.month = anchor[0]
         self.day = anchor[1]
-        self._list_target = []
-        self._list_precursor = []
+        self._target = []
+        self._precursor = []
         self._total_length_target = 0
```
BSchilperoort, line 322: I think the plural `_targets` and `_precursors` would be better.

BSchilperoort (approved):
Hi Yang, thanks for all the changes.

Before you merge I have some small comment. Additionally, I think there are some prospector issues ;)

s2spy/time.py
geek-yang: Good spotting! Corrected!

geek-yang:
Thanks for the final check. Just fix all errors reported by prospectors. I will merge this PR.

PR #110, opened 2022-09-29
Update of 'calendar_development' from 'main'.

BSchilperoort:
The main branch has been updated to include the interactive visualizations with bokeh. As we are working on the development of the CustomCalendar, these will be useful to already merge now so we can make use of them.


geek-yang (approved):
Approved! Now we can review our calendars visually and interactively 😆, feels like from stone age to modern world.

Issue #114, opened 2022-10-04
Changing the `anchor` kwarg of the CustomCalendar


BSchilperoort:
Right now the input format of the `anchor` kwarg is a Tuple(int, int). Where the first int is the month number, and the second int is the day of month.

```py
cal = CustomCalendar(anchor=(12, 31))  # Anchor date is 31 December
```

This works fine for calendars tied to a certain date, however, calendars like the WeeklyCalendar or NWP Calendar cannot work with this calendar. These calendars need a week number (& day of week).

To this end I propose to change the format of `anchor` to a string that accepts the following formats (following ISO 8601):

- `MM`
    -  `"12"` for December. Will be turned into an anchor of 1 December.

- `MM-DD`
    -  `"12-31"` for 31 December.

- `Www`
    - `"W01"` for the first week of the year.

- `Www-D`
    - `"W01-4"` for the first Thursday of the year.
- long month name
    - `"January"`, `"March"`
- short month name
    - `"Jan"`, `"Mar"`
    
For the last two we can use python's builtin `calendar.month_name` and `calendar.month_abbr` as lookup tables.

PR #116, opened 2022-10-05
Modification to the calendar `anchor` kwarg

BSchilperoort:
This PR changes the `anchor` kwarg of the CustomCalendar according to how I laid it out in #114 

Valid anchor input is now of the form:
- `MM` or `MM-DD` for month and day.
- `Www` or `Www-D` for week number and day.
- **English** long or short names of the months

We decided to only allow English names to make the calendar generating code more compatible across machines and locales.

<hr>

This PR also adds CI to the calendar_development branch.


geek-yang (approved):
Nice work! I test all the formats and they all work nicely. The implementation is also quite elegant. Nothing needs to be changed. 

We would like to instruct the user, what formats do we actually support. But I think it makes more sense to put it in the notebooks when we have examples for all the calendars. So this could be an action point when we finish renovating our calendar and try to make some tutorials for the calendars we have.

BSchilperoort:
> We would like to instruct the user, what formats do we actually support. But I think it makes more sense to put it in the notebooks when we have examples for all the calendars. So this could be an action point when we finish renovating our calendar and try to make some tutorials for the calendars we have.

I have updated the docstring!

PR #117, opened 2022-10-05
Refactor `i_intervals`

geek-yang:
In this PR, we change the naming of `i_intervals`. This will influence several modules and they will be fixed in this PR as well.

- Refactor `i_intervals` according to the new naming rule (precursors with minus sign and targets are positive)
- Update plotting (matplotlib_plot) module
- Update resample module

Changes do not affect train/test split module.


s2spy/time.py
```diff
@@ -478,10 +478,19 @@ def _label_targets(self, intervals: pd.DataFrame) -> pd.DataFrame:
...
+            columns={i: f"{self.n_targets - i - 1}" for i in range(self.n_targets,
+             len(intervals.columns))}
         )
 
+        # rename targets
+        intervals = intervals.rename(
+            columns={i: f"{self.n_targets - i}" for i in range(self.n_targets)}
+        )
+
+        return intervals
```
BSchilperoort, line 492: Hi Yang, quick question; why don't we use an IntervalIndex here? I had changed this to having (string) index names to be able to label the targets, but that's not required anymore.

geek-yang: Well spot! It is corrected now.

s2spy/_resample.py
```diff
@@ -219,7 +224,7 @@ def resample(
         raise ValueError("Generate a calendar map before calling resample")
 
     utils.check_timeseries(input_data)
-    utils.check_input_frequency(mapped_calendar, input_data)
+    #utils.check_input_frequency(mapped_calendar, input_data)
```
Peter9192, line 227: Is there a reason to leave commented code in the PR? Why not remove it?

s2spy/_resample.py
```diff
@@ -33,7 +33,7 @@ def mark_target_period(
     if isinstance(input_data, PandasData):
         input_data["target"] = np.zeros(input_data.index.size, dtype=bool)
         input_data["target"] = input_data["target"].where(
-            input_data["i_interval"] >= calendar.n_targets, other=True
+            input_data["i_interval"] < 0, other=True
```
Peter9192, line 36: Didn't we agree that target is >0?  I don't remember?

Peter9192 (commented):
> Refactor i_intervals according to the new naming rule

Can you summarize the new rule here, or link to another item describing it? This description doesn't help if you don't know/remember what we discussed.

s2spy/_resample.py
geek-yang: Because this check is still valid for all calendars with the argument `freq`, but not for `CustomCalendar`. When we finish rebasing everything on `CustomCalendar`, we need to bring it back.

But good comment! Now I add a more explicit note there, just to remind ourselves:

```py
    # This check is still valid for all calendars with `freq`, but not for CustomCalendar
    # TO DO: add this check when all calendars are rebased on the CustomCalendar
    #utils.check_input_frequency(mapped_calendar, input_data)
```

s2spy/_resample.py
geek-yang: Haha. Yep, this is because `input_data["target"] = np.zeros(input_data.index.size, dtype=bool)`, the boolean for target column was initialized to be `False`. So with `DataFrame.where()` we check if the values are not <0, then we switch them as `True`.

As it is not straight-forward (especially for a quick first bite 😄 ), I refactor it as:

```py
        input_data["target"] = np.ones(input_data.index.size, dtype=bool)
        input_data["target"] = input_data["target"].where(
            input_data["i_interval"] > 0, other=False
        )
```

BSchilperoort (approved):
Hi Yang, thanks for fixing this. I did notice a linter issue, related to the mark_target_period method.
We'll have to rethink if mark_target_period is required.

s2spy/_resample.py
```diff
@@ -31,14 +31,14 @@ def mark_target_period(
             given inputs.
     """
     if isinstance(input_data, PandasData):
-        input_data["target"] = np.zeros(input_data.index.size, dtype=bool)
+        input_data["target"] = np.ones(input_data.index.size, dtype=bool)
         input_data["target"] = input_data["target"].where(
-            input_data["i_interval"] >= calendar.n_targets, other=True
```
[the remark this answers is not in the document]
geek-yang: Ooops, missed this one! Thanks for the spotting! It is fixed now. And yep, we can consider deprecating `mark_target_period` in the PR when we update the resample module.

PR #120, opened 2022-10-10
Add tests for custom calendar

geek-yang:
Add tests for custom calendar (only).

For testing, just run `pytest` on `test_time.py` file.


tests/test_time.py
```diff
@@ -5,20 +5,129 @@
...
+        period = Period(length=10, gap=10)
+        assert isinstance(period, Period)
+
+    def test_target_period(self):
+        target = TargetPeriod(20, 10)
+        assert isinstance(target, TargetPeriod)
+
+    def test_precursor_period(self):
+        precursor = PrecursorPeriod(10, -5)
+        assert isinstance(precursor, PrecursorPeriod)
```
BSchilperoort, line 34: Wouldn't it be good to also test that the properties are saved correctly here?
E.g. `assert precursor.length = 10`. (We'll have to change that to TimeDeltas later but putting this in already could be a good idea).

tests/test_time.py
```diff
@@ -5,20 +5,129 @@
...
+
+
+class TestCustomCalendar:
+    """Test CustomCalendar methods."""
+
+    @pytest.fixture(autouse=True)
+    def dummy_calendar(self):
+        cal = CustomCalendar(anchor="12-31")
```
BSchilperoort, line 42: We should probably put in a test that tests a set of anchor inputs

tests/test_time.py
```diff
@@ -5,20 +5,129 @@
...
+        # create target periods
+        target_1 = TargetPeriod(20)
+        # create precursor periods
+        precursor_1 = PrecursorPeriod(10)
+        # append building blocks
+        cal.append(target_1)
+        cal.append(precursor_1)
+        # map years
+        cal = cal.map_years(2021, 2021)
```
BSchilperoort, line 51: The map_to_data method is not tested here yet

BSchilperoort (changes requested):
Hi Yang, nice work. I just think a few extra tests could be useful.

Additionally, a lot of the tests are failing right now. Shall we just disable the tests that we don't need yet right now? We can put the CustomCalendar tests in a different file, and exclude the rest of the `test_*.py` files from testing. This would then allow us to use the CI a bit more efficiently and re-add and fix tests as we get to that part of the code (e.g. resample, RGDR, etc.).

tests/test_time.py
geek-yang: Good idea. Tests added.

tests/test_time.py
geek-yang: Thanks for the comment. Let's create a separate test (specifically for the new parser) in another PR for the different type of anchor inputs.

tests/test_time.py
geek-yang: Add a test for `map_to_data` now.

geek-yang:
> Hi Yang, nice work. I just think a few extra tests could be useful.
> 
> Additionally, a lot of the tests are failing right now. Shall we just disable the tests that we don't need yet right now? We can put the CustomCalendar tests in a different file, and exclude the rest of the `test_*.py` files from testing. This would then allow us to use the CI a bit more efficiently and re-add and fix tests as we get to that part of the code (e.g. resample, RGDR, etc.).

Thanks for the comments @BSchilperoort. I think it is a good idea to mute the "irrelevant" tests for now and make the CD/CI more useful. Following your suggestion, I simply add a dot in front of the file name in those files. And now we get back green tick finally 😄.

BSchilperoort (approved):
Thanks for making the changes! No more comments from me :)

PR #122, opened 2022-10-14
Rebase BaseCalendar to CustomCalendar

geek-yang:
Replace BaseCalendar with CustomCalendar, which includes following steps:

- [x] Replace BaseCalendar with CustomCalendar
- [x] Refactor AdvantCalendar based on the new BaseCalendar
- [x] Refactor MonthlyCalendar and WeeklyCalendarbased on the new BaseCalendar
- [x] Update (relevant) example notebooks
- [x] Fix all tests in time module


s2spy/_base_calendar.py
```diff
@@ -303,10 +363,60 @@ def visualize_interactive(
...
         return self.get_intervals().stack()  # type: ignore
+
+
+class Period(ABC):
+    """Basic construction element of calendar for defining target period."""
+
+    def __init__(self, length: str, gap: str = "0d", target: bool = False) -> None:
+        self.length = DateOffset(**self._parse_time(length))
+        self.gap = DateOffset(**self._parse_time(gap))
+        self.target = target
```
BSchilperoort, line 382: I think these should be properties, or have setters/getters implemented. The following example is accepted, but will break `calendar.append`:
```py
precursor_1 = s2spy.time.PrecursorPeriod("10d", "10d")
precursor_1.length = "14d"
```

tests/test_custom_calendar.py
```diff
@@ -17,23 +17,17 @@ def interval(start, end, closed="right"):
...
-        assert period.gap == 10
-
     def test_target_period(self):
-        target = TargetPeriod(20, 10)
+        target = TargetPeriod("20d", "10d")
         assert isinstance(target, TargetPeriod)
-        assert target.length == 20
-        assert target.gap == 10        
+        assert target.length == DateOffset(days=20)
+        assert target.gap == DateOffset(days=10)
```
BSchilperoort, line 24: Could you add tests here for the other input types we want to support? E.g. weeks and months?

s2spy/time.py
```diff
@@ -224,20 +241,14 @@ def show(self) -> pd.DataFrame:
...
                 periods labelled.
         """
-        return self._label_targets(self.get_intervals()).applymap(
-            self._interval_as_month
-        )
+        return self.get_intervals().applymap(self._interval_as_month)
 
 
-class WeeklyCalendar(BaseCalendar):
+class WeeklyCalendar(AdventCalendar):
```
BSchilperoort, line 247: The docstring for "anchor" has not been updated in the WeeklyCalendar.

BSchilperoort: The example in the docstring also seems to not be tested (anymore)?!

Try the example:
```
import s2spy.time
calendar = s2spy.time.WeeklyCalendar(anchor=40, freq="1W")
```

s2spy/_base_calendar.py
```diff
@@ -52,34 +54,52 @@ def _get_anchor(self, year: int) -> pd.Timestamp:
...
-                max_lag is set to a high enough number that intervals extend over
-                multiple years. `False` by default, to avoid train/test information
-                leakage.
+            anchor_str: Anchor string in the right formatting.
+
+        Returns:
+            Datetime formatter to parse the anchor into a date.
         """
-        if (max_lag < 0) or (max_lag % 1 > 0):
+        if re.fullmatch("\\d{1,2}-\\d{1,2}", anchor_str):
```
BSchilperoort, line 70: From testing it seems we're missing a check for non-string data. Try giving `anchor=10` to a calendar. It will return the following error:
```py
TypeError: expected string or bytes-like object
```
Which is not very clear to users.

BSchilperoort (changes requested):
Hi Yang, good job! I did find some small bugs, but that's expected with such a big overhaul!

I noticed that the `tutorial_resampling_data` is not updated, and does not work at the moment (anchor definition is incorrect). Can you update that for this PR?

Also, I am not a big fan of splitting the calendar explanations over multiple notebooks. How about just two? One for the Advent/Monthly/Weekly calendars, and one for the CustomCalendar? Or one main one using only the Advent calendar, and then a notebook about alternatives?

geek-yang:
> Hi Yang, good job! I did find some small bugs, but that's expected with such a big overhaul!
> 
> I noticed that the `tutorial_resampling_data` is not updated, and does not work at the moment (anchor definition is incorrect). Can you update that for this PR?
> 
> Also, I am not a big fan of splitting the calendar explanations over multiple notebooks. How about just two? One for the Advent/Monthly/Weekly calendars, and one for the CustomCalendar? Or one main one using only the Advent calendar, and then a notebook about alternatives?

Thanks for the review! 😂 I forgot the `tutorial_resampling_data`. Thanks for noticing this. I will fix it.

And I like your idea about only keeping the adventcalendar notebook and merge the others as one alternative_calendar notebook. I will implement this. 👍

s2spy/time.py
geek-yang: I think the docstring check was not switched on before. I will turn it on and fix the docstrings. 😉

s2spy/_base_calendar.py
geek-yang: I agree. I will add a check.

s2spy/_base_calendar.py
geek-yang: Good idea! I will make them "read-only" attributes with property.

tests/test_custom_calendar.py
geek-yang: Added.

geek-yang:
For the other tests (e.g. test_resample.py), it is not relevant with this one and let's fix them in next PR, where I would like to fix traintest and RGDR.

s2spy/_base_calendar.py
```diff
@@ -81,24 +86,24 @@ def _parse_anchor(self, anchor_str: str) -> Tuple[str, str]:
...
 
     def _append(self, period_block):
         """Append target/precursor periods to the calendar."""
-        if period_block.target:
+        if period_block._target:
             self._targets.append(period_block)
             # count length
             self._total_length_target += (
-                period_block.length.kwds["days"] + period_block.gap.kwds["days"]
+                period_block._length.kwds["days"] + period_block._gap.kwds["days"]
```
BSchilperoort, line 99: Why did you change `length` to `_length` here (same for `gap`). These are still accessible as their are properties, so there is no need to use the `_` versions.

s2spy/time.py
```diff
@@ -172,13 +172,13 @@ def __init__(
             n_targets: integer specifying the number of target intervals in a period.
 
         Example:
-            Instantiate a calendar counting down the quarters (3 month periods) until
+            Instantiate a calendar counting down the quarters (3 month periods) from
```
BSchilperoort, line 175: Well spotted :)

BSchilperoort (approved):
Hi Yang, thanks for the changes. I just have one (small) comment! 👍

s2spy/_base_calendar.py
geek-yang: Hmmm...Good point. I didn't think about it. Normally `length` is used for display and `_length` is used internally, as there are always some further processing related to these protected values. But indeed it is not the case here. I will access them via properties. Thanks for the comment 💯 .

PR #123, opened 2022-10-24
Added input checks and tests for anchor kwarg

BSchilperoort:
This PR adds tests for the anchor keyword argument of the BaseCalendar (and thus its children). The test tests for a set of correct inputs, and also a set of incorrect inputs that should be rejected.

Based on the tests, checks were added to the anchor parser, to make sure that all month-day and week-weekday inputs are valid. (E.g. no 31st of April)

### DIFF
--- a/s2spy/utils.py
+++ b/s2spy/utils.py
@@ -169,3 +169,54 @@ def get_month_names() -> Dict:
         "nov": 11,
         "dec": 12,
     }
+
+
+def check_month_day(month: int, day: int = 1):
+    """Checks if the input day/month combination is valid.
+
+    Months must be between 1 and 12, and days must be within 1 and 28/30/31 (depending
+    on the month).
+
+    Args:
+        month: Month number
+        day: Day number. Defaults to 1.
+    """
+
+    if month in {1, 3, 5, 7, 8, 10, 12}:
+        if (day < 1) or (day > 31):
+            raise ValueError(
+                "Incorrect anchor input. "
+                f"Day number {day} is not a valid day for month {month}"
+            )
+    elif month in {4, 6, 9}:
+        if (day < 1) or (day > 30):
+            raise ValueError(
+                "Incorrect anchor input. "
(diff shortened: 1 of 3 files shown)


s2spy/utils.py
```diff
@@ -169,3 +169,54 @@ def get_month_names() -> Dict:
...
+                f"Day number {day} is not a valid day for month {month}"
+            )
+    elif month in {4, 6, 9}:
+        if (day < 1) or (day > 30):
+            raise ValueError(
+                "Incorrect anchor input. "
+                f"Day number {day} is not a valid day for month {month}"
+            )
+    elif month == 2:
+        if (day < 1) or (day > 28):
```
geek-yang, line 198: (Clearly you've thought about it already) Feb 29th is really tricky for the setup...It is ok to put it this way, to simply block the user from doing tricky things 😂.

tests/test_time.py
```diff
@@ -331,3 +331,50 @@ def test_max_lag_skip_years(self, max_lag, expected_index, expected_size):
...
+    ]
+
+    incorrect_inputs = (
+        10,  # non-str inputs
+        (1, 2),
+        "0",  # month number less than 1
+        "13",  # month number greater than 12
+        "12-0",  # day number less than 1
+        "12-32",  # day number greater than 31
+        "31-12",  # incorrect month/day order
```
geek-yang, line 360: One minor case, the combination of month abbreviation and digits (e.g. "Jan-12") is also not supported.

geek-yang (approved):
Nice work! Only got some small things you may consider for changes.

tests/test_time.py
BSchilperoort: You're right! I have added a test for month name + digit.

s2spy/utils.py
BSchilperoort: Yes I don't think we want to consider leap days as valid input!

PR #124, opened 2022-10-24
Updating `test_resample.py` for the calendar changes

BSchilperoort:
This PR will update the tests for the `resample` functionality, by fixing the existing tests to work correctly with the new calendar system, as well as adding tests for overlapping intervals and multidimensional datasets.


geek-yang (approved):
Looks nice!

PR #125, opened 2022-10-24
Update the visualizations to the new calendar system

BSchilperoort:
This PR updates the visualization routines to the new calendar system.

Additionally, the Matplotlib visualization has been refactored so it uses the Bokeh data dictionary to plot data. This simplifies the code base by only having a single source of data for both Matplotlib and Bokeh.

To validate the results, open the AdventCalendar notebook and try out the default `calendar.visualize` and `calendar.visualize_interactive`. Both now have the same `relative_dates` kwarg, which can either be `True` or `False`


geek-yang (approved):
Good job! I tested it with different calendars in different cases and it works perfectly!

PR #126, opened 2022-10-26
Verify propagation of changes by `BaseCalendar` on `traintest` and `RGDR`

geek-yang:
Verify `traintest` and `RGDR` modules with the new `BaseCalendar`, which includes following tasks:

- Verify and fix `traintest` module
- Fix tests for `traintest`
- Verify and fix `RGDR` module
- Fix tests for `RGDR`

### DIFF
--- a/tests/test_traintest.py
+++ b/tests/test_traintest.py
@@ -20,7 +20,7 @@ def dummy_data():
     y = xr.DataArray(np.random.randn(n), coords=time_coord, name="y")
 
     # Map data to calendar and store for later reference
-    calendar = s2spy.time.AdventCalendar(anchor=(10, 15), freq="180d")
+    calendar = s2spy.time.AdventCalendar(anchor="10-15", freq="180d")
     calendar.map_to_data(x1)
     x1 = s2spy.time.resample(calendar, x1)
     x2 = s2spy.time.resample(calendar, x2)
@@ -33,8 +33,8 @@ def test_kfold_x(dummy_data):
     x1, _, _ = dummy_data
     cv = s2spy.traintest.TrainTestSplit(KFold(n_splits=3))
     x_train, x_test = next(cv.split(x1))
-    expected_train = [2019, 2020, 2021, 2022, 2023]
-    expected_test = [2016, 2017, 2018]
+    expected_train = [2019, 2018, 2017, 2016]
+    expected_test = [2022, 2021, 2020]
     assert np.array_equal(x_train.anchor_year, expected_train)
     xr.testing.assert_equal(x_test, x1.sel(anchor_year=expected_test))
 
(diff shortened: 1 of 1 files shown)


notebooks/tutorial_RGDR.ipynb
```diff
@@ -8,21 +8,52 @@
     "This notebook outlines the current status of the RGDR implementation."
```
BSchilperoort, line 8: <p>Perhaps it would be better to keep this as <code>lag=1</code> As the lag is the (interval) difference between the 1st target and the precursors.</p><p> </p><p>Alternatively we can make users pass <code>i_interval</code> instead to avoid this ambiguity, but then there is still the issue of "what is lag" when having <em>two</em> target periods. See <a href="https://github.com/AI4S2S/s2spy/issues/56" target="_blank">#56</a>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/AI4S2S/s2spy/pull/126/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

s2spy/_plot.py
```diff
@@ -188,8 +188,8 @@ def matplotlib_visualization(
         ax.xaxis.set_major_formatter(formatter)
         ax.set_xlim(
             (
-                calendar.flat.values.min().left - pd.Timedelta(days=14),
-                calendar.flat.values.max().right + pd.Timedelta(days=14),
+                intervals.stack().values.min().left - pd.Timedelta(days=14),
+                intervals.stack().values.max().right + pd.Timedelta(days=14),
```
BSchilperoort, line 192: Did this go wrong somehow? Or is this an efficiency improvement (not recalculating the calendar every time)?

BSchilperoort (commented):
Hey Yang, nice job! 👍  I was just wondering about the change in the use of `lag`, and if this is the right way forward. See my comment in the notebook review. Let's discuss this before you merge.

s2spy/_plot.py
geek-yang: Yep~ Since `calendar.flat` includes the whole time, when displaying only a few anchor years (e.g. `n_years = 3`) with `relative_dates=False`, the intervals will become super small. So with `intervals.stack()`, the display is adjusted to the number of years to be included, which looks much nicer 😎.

notebooks/tutorial_RGDR.ipynb
geek-yang: <p>Thanks for the comment, Bart! Indeed I was a bit hesitant when tweaking this, as the selection of minus value sounds a bit weird. And I think it makes sense to call it "lag 1" or "i_interval -1". The "i_interval" reference system helps a bit for the issue of "what is lag" when multiple target periods are included. </p><p> </p><p>Anyway, let's discuss this.</p><p> </p><p> </p>

geek-yang: Following the discussion, to be consistent and to avoid ambiguity, we only use `i_interval` to refer to the index.

PR #130, opened 2022-11-09
Visualization improvements

BSchilperoort:
This PR handles @Peter9192's comments on visualization changes in the Big PR.

- [x] Instead of having two separate visualization methods, it might be nicer to add `backend=matplotlib` , or `interactive=False`
- [x] As a user, I want to be able to plot the calendar before it is mapped to years. This is when it's most important
- [x] Visualize on 20 days gives a cramped image. Use tight_layout?

<hr>

Changes:

 - Both mpl and bokeh are called through `calendar.visualize`
 - Visualization of calendars without mapping is now possible
 - ~Tight layout is called and~ the default figure size is increased
 - Tests are refactored to make them more concise
 - Fixed bug where a calendar with only `PrecursorPeriod`s would break `visualize`
 - Added support to pass a mpl `ax` object to put the plot in
 - Added a `add_legend` kwarg for mpl use
 - Added `**bokeh_kwargs` to give users control over the bokeh plot


Peter9192:
> what do you mean by "Bokeh import could be nicer."?

Sorry! That's not very constructive indeed. I dotted it down but forgot to elaborate. In fact, I think it's rather neatly done. What put me off initially was the pattern

```py
if bokeh_available():
    do stuff
return None
```

In case the if-condition is false, you will get an exception, which is not handled (so why check at all?) and `return None` is never reached. I'm wondering what you think about the following alternatives:

```py
assert_bokeh_available()
do stuff

# or (this would require returning false instead of raising)
if not bokeh_available():
    raise()
do stuff
```

BSchilperoort:
> Sorry! That's not very constructive indeed. I dotted it down but forgot to elaborate. In fact, I think it's rather neatly done. What put me off initially was the pattern
> 
> ```python
> if bokeh_available():
>     do stuff
> return None
> ```
> 
> In case the if-condition is false, you will get an exception, which is not handled (so why check at all?) and `return None` is never reached. I'm wondering what you think about the following alternatives:
> 
> ```python
> assert_bokeh_available()
> do stuff
> 
> # or (this would require returning false instead of raising in the util)
> if not bokeh_available():
>     raise()
> do stuff
> ```

OK! I understand now. I'll fix this!

s2spy/_base_calendar.py
```diff
@@ -344,47 +345,50 @@ def __repr__(self) -> str:
...
+                return bokeh_visualization(
+                    calendar, n_years, relative_dates, add_yticklabels=mapped
+                    )
+        else:
+            _plot.matplotlib_visualization(
+                calendar, n_years, relative_dates, add_length, add_yticklabels=mapped
+            )
 
-            return bokeh_visualization(self, n_years, relative_dates)
         return None
```
Peter9192: Unnecessary return?

s2spy/_base_calendar.py
```diff
@@ -344,47 +345,50 @@ def __repr__(self) -> str:
...
-
-    def visualize_interactive(
-        self,
-        relative_dates: bool,
-        n_years: int = 3,
-    ) -> None:
-        """Plots a visualization of the current calendar setup using `bokeh`.
+        calendar = copy.deepcopy(self)
+        mapped = calendar._mapping is not None  # pylint: disable=protected-access
+        if not mapped:
```
Peter9192, line 370:
```suggestion
        if calendar._mapping is None:  # pylint: disable=protected-access
```

s2spy/_base_calendar.py
```diff
@@ -344,47 +345,50 @@ def __repr__(self) -> str:
...
+            calendar.map_years(2000, 2000)
```
Peter9192, line 371: Should this be a copy to avoid the side-effect of modifying the original calendar?

s2spy/_base_calendar.py
```diff
@@ -344,47 +345,50 @@ def __repr__(self) -> str:
...
+            if not relative_dates:
+                print("Setting relative dates to True, as the calendar is not mapped yet.")
```
Peter9192, line 373: "calendar is not mapped yet" may be unclear. Perhaps something like: "Setting relative_dates=True. To see absolute dates, first call calendar.map_years or calendar.map_data."

s2spy/_base_calendar.py
```diff
@@ -344,47 +345,50 @@ def __repr__(self) -> str:
...
-            add_length: Toggles if the frequency of the intervals should be displayed.
-                      Defaults to False (Matplotlib plotter only)
             n_years: Sets the maximum number of anchor years that should be shown. By
                      default only the most recent 3 are visualized, to ensure that they
                      fit within the plot.
+            interactive: If False, matplotlib will be used for the visualization. If
+                         True, bokeh will be used.
+            relative_dates: Toggles if the intervals should be displayed relative to the
+                            anchor date, or as absolute dates.
+            add_length: Toggles if the frequency of the intervals should be displayed.
```
Peter9192, line 365:
```suggestion
            display_length: Toggles if the frequency of the intervals should be displayed.
```
or `show_length`?

s2spy/_plot.py
```diff
@@ -160,16 +168,17 @@ def matplotlib_visualization(
                         is aligned by the anchor date, so that all anchor years line up
                         vertically.
         add_length: If the length of every periods should be displayed. Defaults False.
+        add_yticklabels: If the years should be displayed on the y-axis ticks.
     """
-    _, ax = plt.subplots()
+    fig, ax = plt.subplots(figsize=(8,5))
```
Peter9192, line 173: Still a bit cramped: 

output

Legend takes up a lot of space, and the axis has an unnecessarily large extent. Details though, don't spend too long on it.

Peter9192 (changes requested):
Thanks! See inline comments

s2spy/_base_calendar.py
BSchilperoort: Already fixed in the last commit! 👍

s2spy/_base_calendar.py
BSchilperoort: Yes, this copy is made as `calendar.map_years` modifies the calendar in-place.

Peter9192: A sorry I completely missed the copy statement above.. my bad!

s2spy/_plot.py
Peter9192: I like that last proposal. How necessary is the legend anyway? Only for first time users I'd say. Perhaps make it optional?

s2spy/_base_calendar.py
BSchilperoort: I changed it to show_length 👍

Peter9192 (approved):
Thanks for addressing the changes @BSchilperoort . Haven't tested by I trust your due diligence

Issue #132, opened 2022-11-10
Default setting of min_area_km2 


semvijverberg:
As a new user, it can be unexpected that some regions are removed by default. The min_area_km2 argument is now 3000**2 by default. You can get region 1 and 3 and think, where is region 2? 

I also noticed there is a small error in the calculation of the area. I was testing at which area_in_km2 a small region of two grid cells of 5x5 would be removed (demonstrated in [tutorial Notebook](https://github.com/AI4S2S/s2spy/blob/main/notebooks/tutorial_RGDR.ipynb)). A back on the envelop calculation would be that this should roughly occur at the following value:

1 degree at equator is ~111 km, at 47,5 degree north this would be:
111 (km) * np.radians(47,5) ~ 92km . So a 5 by 5 grid cell would roughly have an area of (5*92)**2=211.600 km2. 

However, the current calculation of a single grid cell is 1.91e8 km2.

I checked the area calculation [here](https://www.engr.scu.edu/~emaurer/tools/calc_cell_area_cgi.pl), and the answer should be 208763 km2: 

The correct way of calculating the area would be this, following [this] formula (https://www.pmel.noaa.gov/maillists/tmap/ferret_users/fu_2004/msg00023.html).:

```
lat = data.latitude
dlat = np.abs(data.latitude.values[1] - data.latitude.values[0])
dlon = np.abs(data.longitude.values[1] - data.longitude.values[0])

lat_s = np.radians(data.latitude - dlat/2)
lat_n = np.radians(data.latitude + dlat/2)

spherical_area = (np.pi/180) * RADIUS_EARTH_KM**2 * np.abs(np.sin(lat_s) - np.sin(lat_n)) * np.abs(dlon)
```

Which renders and area of 208763 km2 (exactly matching the calculation of the tool above). 

Hence, this issue highlights two things.
- Correction of the function [spherical_area](https://github.com/AI4S2S/s2spy/blob/40475a6d7a36280d262ce5a8c055b85a351ea266/s2spy/rgdr/rgdr.py#L22-L43)
- Suggestion to change the default value of min_area_km2 to 0. 

Could these changes be incorporated in #111 or do we need to raise a new PR?

geek-yang:
@semvijverberg Thanks for finding these bugs and adding a clear explanation. @BSchilperoort and @jannesvaningen could you please inspect the calculation in our package? I think we shall raise a new PR and add this into the unit tests related to RGDR.

Peter9192:
[...]

PR #133, opened 2022-11-11
Fixed spherical area. Removed RGDR kwarg defaults

BSchilperoort:
In #132 Sem noticed an error in the spherical area calculation, and noticed that RGDR still had (arbitrary) default settings.

This PR:

- Fixes the spherical area calculation
- Adds test for the spherical area calculation
- Removes the RGDR default settings

### DIFF
--- a/s2spy/rgdr/rgdr.py
+++ b/s2spy/rgdr/rgdr.py
@@ -334,20 +334,29 @@ class RGDR:
     ) -> None:
         """Response Guided Dimensionality Reduction (RGDR).
 
-        Dimensionality reduction based on the correlation between precursor field and target timeseries.
+        Dimensionality reduction based on the correlation between precursor field and
+        target timeseries.
 
         Args:
             alpha (float): p-value below which the correlation is considered significant
                 enough for a location to be included in a cluster.
-            eps_km (float): The maximum distance (in km) between two samples for one to
-                be considered as in the neighborhood of the other. This is not a maximum
-                bound on the distances of points within a cluster. This is the most
-                important DBSCAN parameter to choose appropriately.
+            eps_km (float): The maximum distance (in km) between two grid cells for them
+                to be considered to be in the same cluster. This is not a maximum bound
+                on the distances between grid cells within a cluster.
+
+                The minimum appropriate value is equal to the maximum width/height of a
+                grid cell (i.e. the width of the grid cell closest to the equator).
+                The upper bound depends on when cells or clusters would still be
+                considered to be part of the same climate signal (i.e., precursor region).
+
+                Higher values can lead to fewer clusters, but also clusters in which the
+                cells of the same cluster are separated by large geographical distances.
             min_area_km2 (float): The minimum area of a cluster (in square km). Clusters
                 smaller than this minimum area will be discarded.
 
         Attributes:
-            corr_map (float): correlation coefficient map of given precursor field and target series
+            corr_map (float): correlation coefficient map of given precursor field and
+                target series.
             pval_map (float): p-values map of correlation
             cluster_map (U20): cluster labels for precursor field masked by p-values
         """
--- a/s2spy/rgdr/rgdr.py
+++ b/s2spy/rgdr/rgdr.py
@@ -330,7 +330,7 @@ class RGDR:
         self,
         eps_km: float,
         alpha: float,
-        min_area_km2: float | None = None,
+        min_area_km2: Optional[float] = None,
     ) -> None:
         """Response Guided Dimensionality Reduction (RGDR).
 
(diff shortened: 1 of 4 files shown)


s2spy/rgdr/rgdr.py
```diff
@@ -38,7 +38,7 @@ def spherical_area(latitude: float, dlat: float, dlon: float = None) -> float:
 
     lat = np.radians(latitude)
     h = np.sin(lat + dlat / 2) - np.sin(lat - dlat / 2)
-    spherical_area = h * dlon / np.pi * 4
+    spherical_area = h * dlon / np.pi / 4
```
semvijverberg, line 41: Minor comment, but to me writing the following is easier to comprehend:

spherical_area = h * dlon / (np.pi * 4)

semvijverberg (commented):
I think it is good to inform on the typical order of magnitude for min_area_km2 in the docstring. Users might not know by heart that a single 1x1 grid cell has an area of roughly 111**2 km2 (at the equator). 

https://github.com/AI4S2S/s2spy/blob/40475a6d7a36280d262ce5a8c055b85a351ea266/s2spy/rgdr/rgdr.py#L72

semvijverberg:
Thanks for the PR @BSchilperoort!

BSchilperoort:
Hi Sem, thanks for the review! I changed the notation in the spherical_area function.

> I think it is good to inform on the typical order of magnitude for min_area_km2 in the docstring. Users might not know by heart that a single 1x1 grid cell has an area of roughly 111**2 km2 (at the equator).
> 
> https://github.com/AI4S2S/s2spy/blob/40475a6d7a36280d262ce5a8c055b85a351ea266/s2spy/rgdr/rgdr.py#L72

I think this is not the right place to give users a typical order of magnitude. For one it does not inform the user on the functionality of RGDR, but rather gives them background information. Besides that, while the area of a 1x1 degree grid cell is ~12000 km^2 at the equator, at 80 degrees north it is only ~2000 km^2. So any suggestion of an order of magnitude would be quite arbitrary, and greatly depends on what the input data is (resolution wise), the study area, etc.

For example, we also do not give suggestions for `eps_km` or `alpha` in the docstring.

s2spy/rgdr/rgdr.py
```diff
@@ -340,8 +343,8 @@ def __init__(
                 be considered as in the neighborhood of the other. This is not a maximum
```
semvijverberg, line 343: eps_km (float): The maximum distance (in km) between two samples for one to
                 be considered as in the neighborhood of the other. This is not a maximum bound on the distances of points within a cluster. This is the most  important DBSCAN parameter to choose appropriately. The appropriate value scales with resolution (as samples will be closer to each other). With a higher resolution, eps_km can be set to lower values and vice versa. Typical values are on the order of $\approx 10^2$.

semvijverberg: It my intention to communicate that distances_eps is generally roughly between 100 and 500. Distances that are smaller than 100 by 100 (smaller then 1 by 1 degree) will likely result in impractical results. Can we perhaps add a link to the documentation where we add a few figure on what happens at the edge cases and what is reasonable?

semvijverberg (commented):
See comments, curious to hear your expert opinion @BSchilperoort.

s2spy/rgdr/rgdr.py
BSchilperoort: I think an appropriate minimum bound for epsilon is the (maximum) width of a gridcell, i.e.:
$\varepsilon_{min} = \textrm{max}(\Delta\textrm{lon}, \Delta\textrm{lat})$

 The upper bound depends on how you want to group clusters by region. Setting it to 1000+ km can make a lot of sense if you would like to group all points within a certain region together, this would also reduce the chance of having a lot of similar small clusters.

What about:
```
eps_km (float): The maximum distance (in km) between two samples for them to be considered 
    to be in the neighborhood of the other. This is not a maximum bound on the distances of points 
    within a cluster.
    The minimum appropriate value is equal to the maximum width/height of a grid cell (i.e. the width of 
    the grid cell closest to the equator). The upper bound depends on when cells or clusters would
    still be considered in the same geographical region.
    Higher values will lead to fewer clusters, and/or clusters that are discontinuous geographically.
```

semvijverberg: I like where this is going: Maybe a minor change related to the word 'geographical region' because it can be confusing I think. 

```
eps_km (float): The maximum distance (in km) between two grid cells for them to be considered 
    to be in the same cluster. This is not a maximum bound on the distances between grid cells
    within a cluster.
    The minimum appropriate value is equal to the maximum width/height of a grid cell (i.e. the width of 
    the grid cell closest to the equator). The upper bound depends on when cells or clusters would
    still be considered to be part of the same climate signal (i.e., precursor region).
    Higher values can lead to fewer clusters, but also clusters in which the cells of the same cluster are separated by large geographical distances.
```

BSchilperoort:
> See comments, curious to hear your expert opinion @BSchilperoort.

I like the description we have gotten to! I think this makes the functionality of `eps_km` a lot more clear to users.

Issue #134, opened 2022-11-16
Pinning `mypy` version


BSchilperoort:
We use `mypy` to check our static types. However, as mypy is still in a 0.x version, the functionality can with new releases, and raise errors in the CI that have nothing to do with changes made by a certain commit.

It is not fun or helpful to see the ❌ show up because of mypy version changes. I'd suggest to pin the mypy version, and periodically update it to avoid this kind of problem.

Example: [0.991 (latest) contains two breaking changes](https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html)

geek-yang:
That sounds good and reasonable to me. Let's fix the version.

Issue #137, opened 2022-11-18
Specifying start/end day and length of a target/precursor period


BSchilperoort:
In @Peter9192's review of #113, he voiced his desire to be able to specify target and precursor periods by the end date and length;
> e.g. if I want to issue a forecast by the end of may, I want to start appending precursors from 31 may backwards.

However, this leads to some non-trivial issues like:

- How do we deal with leap years? As the periods are unaware (yet) of which year they will belong to, we do not know if it is a leap year or not. A "simple" solution is to just ignore leap days.

- How do we determine which _year_ the date specifies?
  - E.g., if the anchor is June 1st, and you specify an end date of June 5th, does this mean 4 days _after_ or 361 days _before_ the anchor. Does the (default) behavior depend on `target=True`/`target=False`?
  - And what if someone wants to define a date two years ahead of the anchor?

And lastly, do we want to support `start_date`, `end_date`, and `length`, and have the user use 2/3 to define the period?

geek-yang:
> e.g. if I want to issue a forecast by the end of may, I want to start appending precursors from 31 may backwards.

I think most of the use cases with this kind of need will be solved by `AdventCalendar`. Only some edge cases would need `CustomCalendar` for backward counting, I assume. I think it is better to come up with new type of calendars to address these special needs rather than change the `BaseCalendar` to allow for more flexibility.

Peter9192:
> * A "simple" solution is to just ignore leap days.

agree, or don't allow it.

> * E.g., if the anchor is June 1st, and you specify an end date of June 5th, does this mean 4 days _after_ or 361 days _before_ the anchor.

Since s2s usecases are usually interested in predicting (ahead of time), I think this would mean 361 days before.



> * And what if someone wants to define a date two years ahead of the anchor?

In the very beginning I had the thought that you could allow this by setting the periodicity of the calendar. Back then we had the adventcalendar doing something like `n_targets = 365 / target_length`, and that could conceptually very easily be changed to a multiple of 365. The idea came from a visualization that Sem showed, something like this: 

![image](https://user-images.githubusercontent.com/17080502/203990080-afe156cf-edc2-439e-a21f-f0997eef724a.png)

you could drag the start and end but it would always have this cyclic pattern. The length of this cycle can be changed though.

PR #138, opened 2022-11-18
Replaced append(Period) with add_interval(kwargs)

BSchilperoort:
This PR will tackle Peter's comments about the Period classes:

- [x] Perhaps we should hide the Period class from the public API. Why not `calendar.add_target_period(...)` and `add_precursor_period`? I found myself doing `calendar.append(PrecursorPeriod("1d"))`, which I think is a sign of suboptimal API
- [x] Relatedly, should it be `add_target`, `add_target_period`, `add_target_interval`, `add_period(target=True)`, `add_interval(target=True)`, or `add_block(target=True)`? What's the difference between a precursor and targetperiod?
- [x] ~I would love to be able to add a period by specifying a length and end day, e.g. if I want to issue a forecast by the end of may, I want to start appending precursors from 31 may backwards.~ See #137 

**Changes:**
 - Intervals are appended using `calendar.add_interval("target", "10d", gap="2d')`
     - TargetPeriod and PrecursorPeriod classes are still used in the backend.
 - Support for non-day lengths and gaps is added, required a rewrite of `skip_nyears`
   - E.g., `calendar.add_interval("target", "2M")`
 - Tests and notebook have been updated
 - `mypy` version is pinned, see #134.


tests/test_custom_calendar.py
```diff
@@ -134,6 +126,17 @@ def test_map_to_data(self, dummy_calendar):
...
+        cal = CustomCalendar(anchor="December")
+        cal.add_interval("target", "1M")
+        cal.add_interval("precursor", "10M")
+        cal.map_years(2020, 2020)
+        expected = np.array(
+            [interval("2020-02-01", "2020-12-01", closed="left"),
+             interval("2020-12-01", "2021-01-01", closed="left")]
+        )
+        assert np.array_equal(cal.flat, expected)
+
```
geek-yang, line 139: Would be nice to have one more test for the `allow_overlap` case.

geek-yang (commented):
Looks really nice to me. I checked the codes and the notebook. The tasks are clearly defined and the implementations are quite clean. Only got one minor comment, which concerns the test. I would wait for the feedback from @Peter9192, before this PR can be merged. Nice work!

Peter9192:
closes #134

Peter9192:
> * Intervals are appended using `calendar.append("target", "10d", gap="2d')`

`calendar.add_interval`, right?

Peter9192 (approved):
Nice!

BSchilperoort:
> > * Intervals are appended using `calendar.append("target", "10d", gap="2d')`
> 
> `calendar.add_interval`, right?

Yes! I mistyped that.

PR #139, opened 2022-11-24
Reworked label alignment into RGDR

BSchilperoort:
This PR implements label alignment over splits into RGDR. The label alignment is aimed at giving similar clusters over different splits the same name, while not changing the actual data (avoiding any train-test leakage).

An example of the final result can be visualized using the following plot:
image

Label alignment is performed by the user as follows:
`s2spy.rgdr.label_alignment.rename_labels(rgdrs, clustered_data)`
Where `rgdrs` is a list of RGDR objects, and `clustered_data` is data that has been clustered using that RGDR object. This can be both the train data as well as the test data.

A new notebook has been added, `example_label_alignment.ipnb" which walks through the train test splitting as well as the label alignment steps.


BSchilperoort:
CI seems borked.

Peter9192:
https://www.urbandictionary.com/define.php?term=borked

Cool, didn't know that word ^^

Issue #142, opened 2022-12-02
Reworked `resample` is slow on larger datasets


BSchilperoort:
Due to the rewrite of the resampling functionality, which was required to support overlapping intervals, the resample operation can be quite slow on larger datasets. 

With some testing I've found a method to speed up the resampling by a factor 50 (!)

With `input_data` being a Dataset with the dims (time: 14957, x: 40, y: 40).

`means_matrix` is a DataArray with a 2-D array with the shape (time: 14957, anchor_interval: 2080). anchor_interval is a stacked dim consisting of all combinations of (anchor_year, i_interval).

**Slow** (~35 s):
```py
nonstacked_resample = xr.dot(input_data["temperature"], means_matrix, dims=["time"])
```

**Fast** (~0.6 s):
```py
input_data_stacked = input_data.stack(xy=("x", "y"))
stacked_resample = np.matmul(means_matrix.values, id_stacked["temperature"].values)
```

These two methods will return the same answer, down to a relative tolerance of 1e-14
```py
# Passes:
np.testing.assert_allclose(stacked_resample,
                           nonstacked_resample.stack(xy=("x", "y")).values,
                           rtol=1e-14)
```

This refactoring can be implemented once the calendar-dev branch is merged to main.

Additionally, if we want to optimize for memory use, we could try to implement this in [sparse](https://sparse.pydata.org/en/stable/index.html).

BSchilperoort:
Fixed in #145

PR #145, opened 2022-12-08
Optimized resample

BSchilperoort:
Implements #142

Summary:
- Resampling should be fast again 😄 
- Reduces total CI time by a factor 2!
- Makes the (xarray) resampling code a bit more complex 👎 

How it works:
- Find vars that should be resampled (data variables with "time" dimension)
- Stack those variables into a single dimension (making it a 2-D array)
- Apply matrix multiplication, which should be fast as it's a multiplication of 2 2-D matrices
- Merge everything back together (to keep the other variables and attributes)

TBD later:
- ~~(try out) [sparse](https://sparse.pydata.org/en/stable/index.html) implementation.~~ Calculation of the matrix is fast, we don't need to store it.
- ~~Support~~ Implement different resampling strategies.

### DIFF
--- a/s2spy/_resample.py
+++ b/s2spy/_resample.py
@@ -157,25 +157,40 @@ def resample_dataset(calendar, input_data: xr.Dataset) -> xr.Dataset:
     data = data.to_dataset()
     data = data.stack(anch_int=("anchor_year", "i_interval"))
 
-    means_matrix = xr.DataArray.from_dict(
-        {
-            "coords": {
-                "time": {"dims": "time", "data": input_data["time"].values},
-            },
-            "dims": ("anch_int", "time"),
-            "data": create_means_matrix(
-                data["interval"].values, input_data["time"].values
-            ),
-        }
+    # Separate data with time dims (should be resampled), from data without time dims
+    #   (which does not need resampling). Otherwise stacking ALL dims together will
+    #   cause the stacked dimension become needlessly large, making resampling slow.
+    input_data_time = input_data[
+        [var for var in input_data.data_vars if "time" in input_data[var].dims]
+    ]
+    input_data_nontime = input_data[
+        [var for var in input_data.data_vars if "time" not in input_data[var].dims]
+    ]
+
+    stacking_dims = list(input_data_time.dims.keys())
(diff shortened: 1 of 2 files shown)


BSchilperoort:
@Peter9192 I have (re-re-re)-refactored the resampling so that we can implemented different methods (median/max/quantile, etc.). However, I have decided to not implement the interface yet in this PR to keep it clean and concise.

s2spy/_resample.py
```diff
@@ -157,25 +157,51 @@ def resample_dataset(calendar, input_data: xr.Dataset) -> xr.Dataset:
...
+    resampled_vars = [xr.DataArray] * len(input_data_time.data_vars)
+    for i, var in enumerate(input_data_time.data_vars):
+        size_input = input_data_time[var].values.shape[1] if stacking_dims else 1
+        resampled_data = np.zeros((size_input, contains_matrix.shape[0]))
+
+        # Note: while looping, contains_matrix will have a size of
+        # (n_intervals * n_anchor_years), making it the most reliably small dim.
+        for j, row in enumerate(contains_matrix):
+            # Note: next line allows for np.median / np.max etc. To be implemented later
+            resampled_data[:, j] = np.mean(input_data_time[var].values[row], axis=0)
```
Peter9192, line 192: Now you're extracting `input_data_time[var].values` `j` times. Would it be faster still if you factor this out of the loop? 

Also: have you experimented with the loop order? Intuitively current order makes sense, but I'd need to refresh my memory on loop optimization to be sure.

BSchilperoort: I tried your suggestion, it makes it very slightly faster (~1%).

I did now realize I should have rewritten it to `resampled_data[j, :]` as numpy is row-major. Refactoring this resulted in a >10% speed increase.

About the loop order: we are currently looping through each row of the "contains_matrix"; which represents each `pd.Interval` of the calendar. This will generally be the smallest dimension in bigger datasets (as it doesn't care about extra dims such as lat, lon, etc.). More importantly, this gives us the data to resample, as I don't think we could even easily loop over the columns of `contains_matrix` for this.

PR #146, opened 2022-12-13
A notebook showcasing all the features of the (new) calendar system

BSchilperoort:
This PR adds a notebook in which all features of the (custom) calendar are showcased. This could be a good starting point for people who are new to either s2spy, or the possible future calendar/time package.


Peter9192 (approved):
Very nice walkthrough!! Few comments:

- In GH preview the figure doesn't seem to render
- Oneliners like `import matplotlib.pyplot as plt; fig = plt.gcf(); fig.autofmt_xdate()` might not be what we want to encourage.
- Should we explain creating overlap by setting a negative gap?
- Do you want to mention there's also an interactive plot?
- Typo in "If you want to to funkier "
- Should we mention that there's also a shorthand class for a monthly and weekly calendar?
- End with a summary line, something like: "Now you are ready to build your own calendar. We're curious to know what you come up with".

BSchilperoort:
> Very nice walkthrough!! Few comments

Thanks for the quick look @Peter9192 ! I did notice the image was broken on GH, should be fixed now.

> Oneliners like import matplotlib.pyplot as plt; fig = plt.gcf(); fig.autofmt_xdate() might not be what we want to encourage.

I did that mostly for when I was working on demo prep. It's a workaround for the dates not rendering nicely. the `fig.autofmt_xdate()` should probably just be incorporated in the plotting utility code.

> Should we explain creating overlap by setting a negative gap?

That's a good idea.

> Do you want to mention there's also an interactive plot?

I guess mentioning it doesn't hurt. Rendering the interactive plot inside the notebook will note be compatible with the online viewer though, so users would have to try the interactive plots on their own machines.

> Should we mention that there's also a shorthand class for a monthly and weekly calendar?

I think it is best to focus on the custom calendar here. I am also feeling more and more like the advent/weekly/monthly calendars are almost obsolete due to the ease of use of the new calendar.

Do we need an entire MonthlyCalendar class, if the following does the same (and can be more flexible as it supports gaps, etc.)
```py
cal = Calendar(anchor="December")
cal.add_interval("target", "1M")
cal.add_interval("precursor", "1M", n=11)
```

PR #152, opened 2022-12-19
First basic preprocessing module

BSchilperoort:
@jannesvaningen and @semvijverberg have a look! The workings of the module are demonstrated in the `tutorial_preprocessing` notebook

### Currently implemented are:

A preprocessor class, which (on .fit):

1. Calculates the rolling mean of the data
2. Computes and stores the (linear) trend of the rolling-mean data.
3. Computes and stores the climatology of detrended rolling-mean data

When calling .transform

1. The input data is detrended using the stored trend
2. The climatology is subtracted from the detrended data

### Example: 
```py
from s2spy import preprocess
preprocessor = preprocess.Preprocessor(
    rolling_window_size=25,
    detrend="linear",
    remove_climatology=True,
)

processed_data = preprocessor.fit_transform(input_data)
```


Peter9192 (commented):
Hey @BSchilperoort, nice that you made a start with this. Overall I like where this is going, but in terms of design, I was hoping for something more composable. 

My main issue is that the preprocessor class is now a "catch-all" object. You already see that it requires a range of input arguments related to different pre-processing steps. If we extend this later on with different operations it will be even more. I would much prefer to have a separate class for each pre-processor task, and then for the overall pre-processor to tie the tasks together. E.g.

```py
class PreprocessorTask(ABC):
    """Interface for the preprocessor tasks.

    Should have fit, transform, and fit-transform.
    """

class Detrend(PreprocessorTask):
    """Implementation of detrend preprocessor."""

class Preprocessor():
    """Pipeline or workflow object that can execute the tasks"""

    def __init__(self, *tasks: PreprocessorTask):
        self.tasks = [task for task in tasks]

    def fit(self, data):
        """Call fit on the each of the subtasks, passing the data on from one task to the next."""

et cetera
```

BSchilperoort:
> Overall I like where this is going, but in terms of design, I was hoping for something more composable.

Thanks for your comments @Peter9192 . Yang did have something akin to your example in mind, and it would be quite easy to implement. 

However, I have not seen the need for this yet. Currently there are 4 input arguments for the entire preprocessor, which is sufficient to cover all we need. The rolling mean is always applied first by default, as @semvijverberg said this is best practice that users should do anyway. Note that the rolling mean is *not* applied when calling .transform.

I have not heard of any additional preprocessors that would need to be implemented (ones that need a .fit method at least), and preprocessing such as decadal/monthly sums can't have a fit method, just transform. At that point we would just be making a wrapper for xarray built-in methods.

Peter9192:
> I have not heard of any additional preprocessors that would need to be implemented

What about dimensionality reduction (RGDR or something different) or feature extraction

> At that point we would just be making a `method.transform` wrapper for xarray built-in methods.

I think this is exactly what we want. It will give us the possibility to control the things that can or cannot be set, and how they are implemented. With xarray you can do almost anything, and you can easily do it "wrong". I thought the idea for us was to provide a consensus/best-practice implementation, with a higher-level (more constrained) interface.

Also, we want things to be modular, right? Putting everything together sounds to me like you're addressing a single use case.

semvijverberg:
Thanks for setting this up. 

*I may not completely follow the reasoning of every detail in the conversation, but I wanted to share some thoughts I have. 

First some jargon. With pre-processing, I and many climate scientists usually refer to only detrending and deseasonalizing, and not e.g. feature extraction methods. We often call those dimensionality reduction methods or clustering methods and I suggest that we want to stick to that convention. Pre-processing can also refer to handling NaNs and interpolation methods. Roughly speaking, it should encompass preparing the data before you feed it to (the first step of) your analysis. 

- The rolling mean is the most computational heavy job for the pre-processing. You want to avoid doing that twice, which I believed was the motivation to put everything under one class. 
- Like @BSchilperoort mentioned, (in principle) the deseasonalizing and detrending should be done on rolling mean (smoothened) data (if data is daily/weekly). I thought I also did my detrending on the smoothened data, but I recently discovered I did not do this, oeps. Anyway, I still think we should detrend on smoothened data, but there should be some visual verification on the implications (raw vs smoothened) and checking if it works as desired. 
- I think you know this but to be sure, an awesome new feature is that we allow the detrending/deseasonalizing to be done using only the training data.

BSchilperoort:
Thanks for your input, @semvijverberg. I did have an talk with Peter before the holidays where we discussed why I implemented it this way. I think we're all mostly on the same page now. 

One point Peter did raise, which I do agree with, is that the current name `Preprocessor` is not ideal. As you yourself say:
> Pre-processing can also refer to handling NaNs and interpolation methods. Roughly speaking, it should encompass preparing the data before you feed it to (the first step of) your analysis.

In the same sense, resampling to the Calendar system is also a form of preprocessing. 

Would you (or @jannesvaningen ) know of a good name to give this detrending+deseasonalizing step (instead of the current generic `Preprocessor`)?

semvijverberg:
> Thanks for your input, @semvijverberg. I did have an talk with Peter before the holidays where we discussed why I implemented it this way. I think we're all mostly on the same page now.
> 
> One point Peter did raise, which I do agree with, is that the current name `Preprocessor` is not ideal. As you yourself say:
> 
> > Pre-processing can also refer to handling NaNs and interpolation methods. Roughly speaking, it should encompass preparing the data before you feed it to (the first step of) your analysis.
> 
> In the same sense, resampling to the Calendar system is also a form of preprocessing.
> 
> Would you (or @jannesvaningen ) know of a good name to give this detrending+deseasonalizing step (instead of the current generic `Preprocessor`)?

Yes, this is an excellent point. To be honest, I have no good answer. Perhaps simply class detrend_deseasonalize?

Peter9192:
> > Would you (or @jannesvaningen ) know of a good name to give this detrending+deseasonalizing step (instead of the current generic `Preprocessor`)?
> 
> Yes, this is an excellent point. To be honest, I have no good answer. Perhaps simply class detrend_deseasonalize?

What about "normalize" or "standardize" perhaps in conjunction with "TimeSeries", or a shortened "TS". Classnames could be `TSNormalize` or `StandardizeTS`, for example.

jannesvaningen:
I am currently struggling with the same type of issues while adding functionality to the legacy code for out-of-sample preprocessing... I think you want to give the user the option to do one, two or all of the three methods that we mentioned, but only in a specific order:

1. rolling mean
2. deseasonalize
3. detrend

The user would have the freedom to do only detrending, but not do detrending and then do deseasonalizing. The aim here is to maintain the best practices. I have no idea how you could best accompany this in code. Do you agree? And if so, any ideas?

For the naming: in econometrics, estimating a trend, cycle or seasonality in a timeseries is called 'timeseries decomposition'. But this only concerns the 'fit' part of this exercise, not the 'transform' part. As far as I know, there is no good general word for all of the transform steps (detrend, deseasonalize) since they are considered separately (if there is no seasonality, why deseasonalize?). What @Peter9192 suggests comes close, but standardization or normalization can also be confusing because these are used in a context where the ts is scaled compared to its standard deviation.

If you follow all the steps you are left with a timeseries that should be stationary or integrated of order 0. The procedure to get such a timeseries is referred to as _differencing_.  

So maybe `TSDecompose` or `TSDifferencing`?

Peter9192:
> The user would have the freedom to do only detrending, but not do detrending and then do deseasonalizing. The aim here is to maintain the best practices. I have no idea how you could best accompany this in code. Do you agree? And if so, any ideas?

I think this should not be too difficult. The interface (what it looks like to the user) could be very different, but here's an example implementation:

```py

class Workflow:
    def __init__(self):
        self._tasks = {}

    def add_task(self, task):
        self._tasks[task.name] = task

    def execute(self):
        order = ['A', 'B', 'C']
        for task in order:
            if task in self._tasks:
                self._tasks[task].execute()

class Task:
    def __init__(self, name):
        self.name = name

    def execute(self):
        print(f"Executing {self.name}")

workflow = Workflow()
workflow.add_task(Task('C'))
workflow.add_task(Task('B'))
workflow.execute()
> Executing B
> Executing C
```

semvijverberg:
> > > Would you (or @jannesvaningen ) know of a good name to give this detrending+deseasonalizing step (instead of the current generic `Preprocessor`)?
> > 
> > 
> > Yes, this is an excellent point. To be honest, I have no good answer. Perhaps simply class detrend_deseasonalize?
> 
> What about "normalize" or "standardize" perhaps in conjunction with "TimeSeries", or a shortened "TS". Classnames could be `TSNormalize` or `StandardizeTS`, for example.

I agree with @jannesvaningen, words like "normalize" and "standardize" refers to scaling timeseries. Removing the seasonal cycle is definitely different.

semvijverberg:
> If you follow all the steps you are left with a timeseries that should be stationary or integrated of order 0. The procedure to get such a timeseries is referred to as _differencing_.
> 
> So maybe `TSDecompose` or `TSDifferencing`?

TSDecompose is also used by [statsmodels](https://www.statsmodels.org/dev/generated/statsmodels.tsa.seasonal.seasonal_decompose.html#statsmodels.tsa.seasonal.seasonal_decompose). They use the word 'decompose' referring to both quantify the trend + seasonal cycle within one function. Btw, I think the statsmodels implementation is in not well documented. I'm not sure what happens under the hood.  


TSDifferencing is a well-known approach to create a stationary timeseries (no trend, stationary variability), but it is something very different from what we are doing. I find the method always a bit funky and not physically logical.

s2spy/preprocess.py
```diff
@@ -0,0 +1,220 @@
...
+            data: Input data to preprocess.
+
+        Returns:
+            data_rolling: Same as input but with rolling mean applied.
+        """
+        _check_input_data(data)
+
+        data_rolling = data.rolling(
+            dim={"time": self._window_size}, min_periods=self._min_periods, center=True
+        ).mean()
```
geek-yang, line 145: It should be noticed that the `xarray.rolling` function does not modify data in place. So we should return it and use the return data for `transform`.

s2spy/preprocess.py
```diff
@@ -0,0 +1,220 @@
...
+            self._climatology = _get_climatology(
+                (
+                    _apply_trend(data_rolling, self._detrend, self._trend)
+                    if self._detrend is not None
+                    else data_rolling
+                )
+            )
+        self._is_fit = True
+
+        return data_rolling
```
BSchilperoort, line 162: @geek-yang `.fit` should not return data. Additionally, the rolling mean should only be used to calculate the climatology (and can be used for the trend, but that doesn't really matter). 
A rolling mean should not be applied to data with `.transform`.

See the docstring I wrote for the __init__ method.

BSchilperoort: We should also not store the rolling mean (just use it for the climatology, then discard it), to avoid unnecessary memory use.

geek-yang: Thanks for your explanation @BSchilperoort . But if so the user cannot get the rolling mean data during preprocessing. There will be cases that the user want rolling mean rather than the raw data (e.g. in my work related to energy transport, I need to take the rolling mean first and use it since then. The rest of the work has nothing to do with the raw input since it is too noisy), therefore I think it is necessary to provide this option. We can talk about it in next meeting.

geek-yang:
Just performed a test, it shows that by making the rolling window to be 1, it still performs rolling mean calculation, which is not computationally efficient. We should skip the rolling mean operation when the window is set to be `None`.

s2spy/preprocess.py
geek-yang: To conclude from what we've discussed today, we should not return anything from `fit` as it breaks the pipeline. Rolling mean should either be handled by another function specific for rolling mean with `fit` and `transform`, or by the users themselves (as it is just a simple operation within the xarray). Sorry for the confusion. The changes are reverted now.

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+
+        When calling `.transform(data)`, the preprocessor will:
+         - Remove the (stored) trend from a copy of the data.
+         - Remove the climatology from this detrended data.
+         - Return the detrended and deseasonalized data.
+
+        Args:
+            rolling_window_size: The size of the rolling window that will be applied
+                before calculating the trend and climatology. Setting this to 1 will
+                effectively skip this step.
```
BSchilperoort, line 114:
```suggestion
            rolling_window_size: The size of the rolling window that will be applied
                before calculating the trend and climatology. Setting this to None will
                skip this step.
```

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+        self._is_fit = False
+
+    def fit(self, data: Union[xr.DataArray, xr.Dataset]) -> None:
+        """Fit this Preprocessor to input data.
+
+        Args:
+            data
+        """
+        _check_input_data(data)
+        if self._window_size is not None:
```
BSchilperoort, line 139:
```suggestion
        if self._window_size not in [None, 1]:
```
Perhaps we should also just write a check to make sure that the window size is >=1.

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+
+    Args:
+        data: The input data of which you want to know the trend.
+
+    Returns:
+        Dictionary containing the linear trend information (slope and intercept)
+    """
+    slope, intercept = xr.apply_ufunc(
+        _linregress,
+        data["time"].astype(float),
```
jannesvaningen, line 35: this astype(float) is a super nice solution! It easily allows to apply the trend to another (out-of-sample) dataset. This is something we were really struggling with @semvijverberg

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+        input_core_dims=[["time"], ["time"]],
+        output_core_dims=[[], []],
+        vectorize=True,
+    )
+    return {"slope": slope, "intercept": intercept}
+
+
+def _apply_linear_trend(data: Union[xr.DataArray, xr.Dataset], trend: dict):
+    """Apply a previously calclulated linear trend to (new) data."""
+    return data - trend["intercept"] - trend["slope"] * (data["time"].astype(float))
```
jannesvaningen, line 46: But what if the applied data does not have the same time axis format? So if the new data has days since 1900 or so, while the trend was calculated on nanoseconds since 1900-01-01 00:00:00 ?

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def _get_trend(data: Union[xr.DataArray, xr.Dataset], method: str):
+    """Calculate the trend, with a certain method. Only linear is implemented."""
+    if method == "linear":
+        return _trend_linear(data)
+    raise ValueError(f"Unkown detrending method '{method}'")
+
+
+def _apply_trend(data: Union[xr.DataArray, xr.Dataset], method: str, trend: dict):
```
jannesvaningen, line 56: Can we think of a different name? Apply trend sounds like we are adjusting the trend towards a newly calculated one. Maybe remove_trend, similar to remove_climatology?

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+        )
+    if "time" not in data.dims:
+        raise ValueError(
+            "Analysis is done of the 'time' dimension, but the input data"
+            f" only has dims: {data.dims}"
+        )
+
+
+class Preprocessor:
+    """Preprocessor for s2s data."""
```
jannesvaningen, line 91: It is not really only for 's2s' data, right? More generally for climate data?

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+            data
+
+        Returns:
+            Preprocessed data.
+        """
+        if not self._is_fit:
+            raise ValueError("The preprocessor has to be fit to data before a transform"
+                             " can be applied")
+
+        if self._detrend is not None:
```
jannesvaningen, line 175: Really nice and concise! Although the class docstring clearly describes the steps and the order, I think it would be nice to give some more user feedback on which steps were taken in what order. Now you just specify if you want to do rolling mean or detrend etc. but if I call transform it is not clear to me in what order the steps were taken. In our company code, we provide a plotting option for every step to show what happens during the transform like so: 
image1
It could also be some print statements like: detrending... or deseasonalizing.. just to know what is happening without looking at the code.

s2spy/preprocess.py
```diff
@@ -0,0 +1,217 @@
...
+def _apply_trend(data: Union[xr.DataArray, xr.Dataset], method: str, trend: dict):
+    """Apply a previously calculated trend to (new) data. Only linear is implemented."""
+    if method == "linear":
+        return _apply_linear_trend(data, trend)
+    raise NotImplementedError
+
+
+def _get_climatology(data: Union[xr.Dataset, xr.DataArray]):
+    """Calculate the climatology of timeseries data."""
+    return data.groupby("time.dayofyear").mean("time")
```
jannesvaningen, line 65: What happens if the data is not daily?

BSchilperoort: That's still a ToDo. We have an item open on our dev board for this.

s2spy/preprocess.py
semvijverberg: Note, if you apply both trend and intercept, you will remove the mean value of the timeseries and therefore not only remove the trend but also center the value around 0 (which I would not want as a user).

jannesvaningen:
Nice work Bart! I added some comments, hope it wasn't too much. I added them as single comments, looks a bit messy now, sorry for that. I love how concise and elegant this preprocessor is. And it works fast and smoothly!

I know this is only the first basic preprocessing module, but I hope we can remind ourselves of all the steps that were described in the preprocessing discussion. I would love to also see implemented:
- spatial selection (so selecting a certain spatial extent with lat and lon)
- regridding
- apply land/sea mask

s2spy/preprocess.py
semvijverberg: In principle I did a similar thing, but then the floating number was always anchored at 1970-01-01 (this is following the UNIX time definition).

s2spy/preprocess.py
BSchilperoort: I think that plotting modules do not fit well in this simple preprocessing module. The operations are so simple than it is quite trivial for a user to apply them separately and/or plot them before and after transformation, for example:
```py
data_array.plot()
data_array = preprocessor.transform(data_array)
data_array.plot()
```

As for print statements, they do not belong in (non-debugging) code. [You won't find a single one in sklearn, for example](https://github.com/scikit-learn/scikit-learn/search?q=print%28). If a user wants to know what is going on, they can look at:

- typing `help(function_name)`
- follow examples in the documentation
- look at the auto api reference
- (as a last resort) look at the code

BSchilperoort:
> Nice work Bart! I added some comments, hope it wasn't too much. I added them as single comments, looks a bit messy now, sorry for that. I love how concise and elegant this preprocessor is. And it works fast and smoothly!

Thanks for the review! Don't worry about the single comments. It's only a little bit less compact than a review with multiple comments.

s2spy/preprocess.py
BSchilperoort: `_subtract_trend` is probably a better name, yes. It is probably also more accurate to rename the other one to `_subtract_climatology`.

s2spy/preprocess.py
BSchilperoort: > center the value around 0 (which I would not want as a user).

Why would you not want that? You will be left with the anomaly. Otherwise you are left with `anomaly + mean`.

To me it feels quite arbitrary to add the mean value at, for example the middle year, back to the data. Why not choose the earliest year, or the latest year? Or always a specific year.

I believe that "detrending" basically always removes the mean by default. The "trend" consists of the "slope" + "intercept".

s2spy/preprocess.py
BSchilperoort: > but then the floating number was always anchored at 1970-01-01 

That is exactly what this does, assuming the format of data["time"] is a np.datetime64, which is how timestamps are stored in xarray data.

We should probably add a check that the time coordinate does indeed have timestamps.

s2spy/preprocess.py
semvijverberg: Yes, I agree that the classical statistical definition is like that, but as a scientist manipulating the data this is not what I want.  

We linearly detrend as a proxy for the climate change signal. The intercept is both the mean over all timesteps + the offset due to the trend (e.g., a positive trend coefficient * a positive x-value results in a mean positive offset). In my view, removing the trend-line means something physical, but removing the intercept is more like statistical artifact of the analysis. If you would want to get anomalies, you would need to remove the climatology (which is calculated differently).

s2spy/preprocess.py
geek-yang: Or to be even more general, spatial-temporal data.

s2spy/preprocess.py
geek-yang: Just create an issue to keep tracking it #160.

s2spy/preprocess.py
geek-yang: Just notice this. Yes, I think in meteorology we only want to remove the trend not the intercept. I cannot even think about a usecase for climate study in which the intercept is removed together with the trend. As what @semvijverberg said, if you want to get anomalies (which are centralized around zero), we will remove the climatology. Let's remove the intercept in this function.

geek-yang:
> Nice work Bart! I added some comments, hope it wasn't too much. I added them as single comments, looks a bit messy now, sorry for that. I love how concise and elegant this preprocessor is. And it works fast and smoothly!

Thanks for your review @jannesvaningen! I think it helps to identify a few issues that could be addressed to improve the preprocessor.

> I know this is only the first basic preprocessing module, but I hope we can remind ourselves of all the steps that were described in the preprocessing discussion. I would love to also see implemented:
> 
> * spatial selection (so selecting a certain spatial extent with lat and lon)
> * regridding
> * apply land/sea mask

Thanks for your suggestions. We've thought about this when designing the preprocessor. But given that the users are asked to prepare their data as data arrays, these steps could be easily addressed by calling the `xarray` (or `xesmf`) methods, for instance:

- spatial selection `da.sel(lat=slice(20, 30), lon=slice(110, 140))`
- for regridding, we can do simple interpolation with `da.interp()` or `da.interp_like()`, and for complex ones with xesmf (e.g. `regridder = xe.Regridder(ds, ds_out, "bilinear")` and then `ds_out = regridder(ds)`)
- apply land/sea mask `da.where(np.isnan(mask))`

I think we agree to keep this preprocessor light weight, so it is not very attractive to me to re-implement all these steps rather than adding a notebook to showcase everything. But it maybe useful to add all these steps in one place (e.g. a function to receive a dict incl. all the preprocessing steps, just like a workflow builder), which makes it easier for the user. I would like to support it in later iterations if necessary, depending on real usecases.

s2spy/preprocess.py
semvijverberg: yes indeed, thanks for the additional info Yang. It is indeed a meteorology/climate science thingy. 

Do note the offset you need to take into account. We also fell into that pitfall in our first attempt. 

I fixed it like this, but there is definitely a more elegant way ;).
```
      # the mean of x-axis * trend is not zero. Determining in-sample offset of trend
      A = np.vstack([x, np.zeros(len(x), dtype="float64")]).T
      trends_offsets = np.dot(A, np.vstack([trends, intercepts])).mean(0)
```

geek-yang:
> Nice work Bart! I added some comments, hope it wasn't too much. I added them as single comments, looks a bit messy now, sorry for that. I love how concise and elegant this preprocessor is. And it works fast and smoothly!
> 
> I know this is only the first basic preprocessing module, but I hope we can remind ourselves of all the steps that were described in the preprocessing discussion. I would love to also see implemented:
> 
> * spatial selection (so selecting a certain spatial extent with lat and lon)
> * regridding
> * apply land/sea mask

A new issue #161 has been created to document these suggestions.

s2spy/preprocess.py
geek-yang: Renaming done.

s2spy/preprocess.py
geek-yang: Corrected!

tests/test_preprocess.py
```diff
@@ -102,6 +139,55 @@ def test_transform_without_fit(self, preprocessor, raw_field):
...
+            preprocessor_no_detrend.trend
+
+    def test_climatology_property_not_fit(self, preprocessor):
+        with pytest.raises(ValueError, match="The preprocessor has to be fit"):
+            preprocessor.climatology
+
+    def test_trend_property_no_climatology(
+        self, preprocessor_no_climatology, raw_field
+    ):
+        preprocessor_no_climatology.fit(raw_field)
```
BSchilperoort, line 191:
```suggestion
```
This step should not be needed right? If `subtract_climatology` is `False` then no matter if the preprocessor is fit or not, it can't give the climatology :)

docs/notebooks/tutorial_preprocessing.ipynb
```diff
@@ -0,0 +1,285 @@
...
+    "    subtract_climatology=True,\n",
+    ")"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The preprocesssor can be fit to data, which changes the object in place:"
```
BSchilperoort, line 53:
```suggestion
    "The preprocesssor can be fit to data, which updates the preprocessor object in place, storing the trend and climatology:"
```
Perhaps we should be a bit more explicit in the notebooks

docs/notebooks/tutorial_preprocessing.ipynb
```diff
@@ -0,0 +1,285 @@
...
+   "source": [
+    "preprocessor.climatology[\"sst\"].isel(latitude=0, longitude=0).plot()"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The trend has multiple components, depending on the method. The \"linear\" method has a slope and an intercept.\n",
```
BSchilperoort, line 143:
```suggestion
    "The trend has multiple components, depending on the method. The \"linear\" method has a slope and an intercept. (Currently, only linear detrending is implemented). \n",
```

BSchilperoort (commented):
Nice! Thanks for making the changes Yang. I have suggested some small modifications 😄 

I also thought it would be nice (and more complete) to expose the stored parameters as `@property`s. E.g.:

```
@property
def detrend(self):
    return self._detrend
```

etc. for the other parameters. This makes it easier to inspect this item later, for users who are for example working in a notebook.

We could also open an issue for this, and then also add a nicer repr which tells you the settings (and if the preprocessor is fit already, with an `is_fit` property).

tests/test_preprocess.py
geek-yang: Hmmm....I just noticed that I forgot to move "no climatology" error in the front of the "not fit" error. Yes, afterward this line is not needed.

geek-yang:
> Nice! Thanks for making the changes Yang. I have suggested some small modifications 😄
> 
> I also thought it would be nice (and more complete) to expose the stored parameters as `@property`s. E.g.:
> 
> ```
> @property
> def detrend(self):
>     return self._detrend
> ```
> 
> etc. for the other parameters. This makes it easier to inspect this item later, for users who are for example working in a notebook.
> 
> We could also open an issue for this, and then also add a nicer repr which tells you the settings (and if the preprocessor is fit already, with an `is_fit` property).

Thanks for your review @BSchilperoort . Yes, I agree that we need to have a `repr` to inform the user what they have created. I just create an issue #162 for it. Let's do it in another PR.

semvijverberg:
Cool! Thanks @BSchilperoort and @geek-yang!

Issue #151, opened 2022-12-19
correlating across both intra-seasonal and inter-seasonal variability ('subseasonal mode')  


semvijverberg:
The current implementation is forcing the RGDR().fit method to correlate across the anchor_year dimension (1 datapoint per year). I would like to flatten the anchor_year and i_interval year in order to correlate once using all datapoints (instead of looping over each target). 

To solve this issue, I propose to add the 'corr_dim' argument to RGDR().fit. Such that people can also use it more flexibly (even without creating a calendar). 

This is what I want to do: 
'''
target = target_resampled['t2m'].sel(i_interval=(slice(1,6))).stack(time=['anchor_year', 'i_interval']) 
field = field_resampled.sel(i_interval=slice(-1, 5)).stack(time=['anchor_year', 'i_interval'])

RGDR().fit(field, target, corr_dim='time')
'''

Any other comments suggestions? I know this is already supported by https://github.com/AI4S2S/s2spy/blob/825d359e9bc02313a97c222f72699993b611a3fb/s2spy/rgdr/rgdr.py#L274-L276 so should be a very minor change.

BSchilperoort:
Hi Sem, I tried out code similar to your example, and I was indeed able to get the right correlation out using `rgdr.correlation`

Modified from `rgdr_tutorial.ipynb`:
```py
target_data = target_resampled.sel(cluster=3).ts.sel(i_interval=(slice(1,6))).stack(anch_int=['anchor_year', 'i_interval'])
field_data = field_resampled.sst.sel(i_interval=slice(-1, 5)).stack(anch_int=['anchor_year', 'i_interval'])

field_data["anch_int"] = range(field_data["anch_int"].size)
target_data["anch_int"] = range(target_data["anch_int"].size)

corr, p_val = correlation(field_data, target_data, corr_dim="anch_int")
```

However, it is essential that:
- the number of intervals in both the target and the field are the same
- the target and field data is properly stacked
- users select the right intervals for both field and target

Perhaps it could be better to modify RGDR, rather than the `.fit` method. As `.fit` should not really take any input other than data. We then can put the interval handling and dim stacking inside RGDR. How about the following syntax:
```py
rgdr = RGDR(
    target_intervals=[1, 2, 3],  #int or list
    lag=2  # cross correlation lag. Would make precursor_intervals=[-2, -1, 1]
)
```

semvijverberg:
Yes definitely agree with all suggestions! Would love that feature!

PR #153, opened 2023-01-03
Support multiple targets/precursors in RGDR

### LINKED ISSUE #56: Supporting multiple target intervals in `map_correlation`.

BSchilperoort:
This PR adds the possibility to specify multiple target periods for RGDR, e.g.:
```py
RGDR(
    target_intervals=[1, 2, 3],
    lag=3,
    ...
)
```
To be able to apply these changes the functionality that allows users to apply RGDR to a whole set of precursor intervals, and then analyze them separately, had to be removed. Specifying the interval indices is now required when initiating `RGDR`, and when calling `.transform` the to-be-transformed data needs to have the right precursor intervals.

- Implements #151 
- Closes #56
- Small improvements in RGDR class (added `@prop`s, `__repr__`, ...)
- Some RGDR preview plot improvements
- Corrects docstring in `s2spy.rgdr.label_alignment.rename_labels()` (based on previous comment of Peter).
- Fixes a typing issue in `s2spy.time.resample` (using `@typing.overload`)


s2spy/rgdr/rgdr.py
```diff
@@ -355,91 +355,132 @@ def __init__(
...
+        )
+
+        self._corr_map: Union[None, xr.DataArray] = None
+        self._pval_map: Union[None, xr.DataArray] = None
+        self._cluster_map: Union[None, xr.DataArray] = None
+
+        self._area = None
+
+    @property
+    def target_intervals(self) -> List[int]:
```
semvijverberg, line 382: For my own learning experience, can I ask what the added value of this function is? self._target_intervals is already a list (defined at __init__), so why create this function and not simply name:

```
        self.target_intervals = (
             [target_intervals]
             if isinstance(target_intervals, int)
             else target_intervals
         )
```

semvijverberg: Maybe it's for consistency reasons? Later on I see the added value of calling the method cluster map for example.

semvijverberg (approved):
Amazing work @BSchilperoort! Thanks! 

Super nicely documented, tested, and coded.

s2spy/rgdr/rgdr.py
BSchilperoort: The `@property` decorator is so we can give users information without allowing them to change anything. It is not a function in the way that users call it like a class property, e.g.:
```py
rgdr.target_intervals
```
And not:
```py
rgdr.target_intervals()  # raises an error
```

We do not want to allow users to change properties of the RGDR class after it is initiated. The underscore at the beginning of the variable name `self._target_intervals` denotes that users _should not_ touch this, and if they do so anyway that they're responsible for any weird behavior.

If we do not implement `@property`, a user would be able to do:
```py
rgdr.target_intervals = 5
```
And the lag & precursor intervals would not be affected.

If it is a `@property`, and no `@target_intervals.setter` has been implemented, Python will raise an error when a user tries to modify the value.

This is also the reason why I made the stored maps properties.

notebooks/tutorial_RGDR.ipynb
```diff
@@ -78,72 +84,92 @@
...
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### The target timeseries comes from already pre-clustered land surface temperature data."
+    "### The target timeseries comes from already pre-clustered land surface temperature data.\n",
+    "RGDR requires `DataArray`s as input, as otherwise it is ambiguous which variable should be used.\n",
+    "\n",
+    "When initializing RGDR, the target interval(s) and lag have to be specified as well.\n",
+    "The precursor interval(s) will be determined from the target intervals and lag.\n",
```
Peter9192, line 104: Add a (more explicit) note here that the maximum lag is constrained by the range of the available precursor intervals?

Peter9192 (approved):
I briefly looked over the code and examples, looks very good! The modified functionality makes sense to me.

I was wondering whether the example illustration for multiple targets also has a source file (for easy adaptation later on). Also, in the example notebook, I think it would be helpful to have a similar explanatory figure for the single target interval.

s2spy/rgdr/label_alignment.py
```diff
@@ -358,15 +357,14 @@ def _rename_datasets(
         A list of the input clustered data, with the labels renamed.
     """
     renamed = []
-    for split, _rgdr in enumerate(rgdr_list):
+    for split, _ in enumerate(rgdr_list):
         # A copy is required to not modify labels of the input data
         data = copy(clustered_data[split])
-        labels = data["cluster_labels"].values
-        i_interval = str(_rgdr.cluster_map["i_interval"].values)  # type: ignore
+        labels = data["cluster_labels"].values.astype("U32")
```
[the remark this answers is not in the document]
BSchilperoort: In xarray you cannot have arbitrarily long strings as coordinates, so I chose for a 32 character Unicode string, as that would fit well enough if users would want to rename clusters to something more descriptive like "north-west pacific ocean".

Otherwise the name would be trimmed if users try to change values in the array. E.g.:
image

notebooks/tutorial_RGDR.ipynb
BSchilperoort: Yes, good point. It is now explicitly mentioned.

BSchilperoort:
> I briefly looked over the code and examples, looks very good! The modified functionality makes sense to me.
> 
> I was wondering whether the example illustration for multiple targets also has a source file (for easy adaptation later on). Also, in the example notebook, I think it would be helpful to have a similar explanatory figure for the single target interval.

Thanks for the review, Peter. Good point about adding the source file (and illustration for the single-target case). I have converted the ppt file to svg files, and added the single-target illustration.

PR #158, opened 2023-02-15
Rebase all to lilio

BSchilperoort:
- [x] Deleted duplicate code
- [x] Import lilio modules
- [x] Update readme
- [x] Fix label alignment flakey test


geek-yang (approved):
Nice work! Just a minor comment.

PS: This repo looks so empty now 😰 ...

notebooks/example_label_alignment_RGDR.ipynb
```diff
@@ -23,21 +23,20 @@
                   "source": [
```
geek-yang, line 23: <p>This looks a bit boring compared to what it was before. But it seems there is no change related to RGDR...looks really strange..</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/AI4S2S/s2spy/pull/158/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

BSchilperoort: Yeah the reason that things changed is because the order of the anchor_years changed. This causes a different split of the data, and as such different clusters

PR #159, opened 2023-02-16
Improve quality of life by using hatch and ruff

geek-yang:
In this PR, we move to hatch and ruff to improve the configuration, testing and linting of our package. Following tasks are expected to be complete in this PR:

- [x] Use hatch
- [x] Use ruff
- [x] Fix linters
- [x] Remove `autouse` in all tests
- [x] Remove unneeded dependency
- [x] Update changelog


pyproject.toml
```diff
@@ -1,10 +1,143 @@
...
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 
+[tool.mypy]
+ignore_missing_imports = true
+
+[tool.black]
+line-length = 88
+src_paths = ["s2spy", "tests"]
+line_length = 120
```
BSchilperoort, line 131:
```suggestion
```

s2spy/rgdr/rgdr.py
```diff
@@ -57,12 +57,7 @@ def cluster_area(ds: XrType, cluster_label: float) -> float:
     Returns:
         float: Area of the cluster `cluster_label`.
     """
-    return (
-        ds["area"]
-        .where(ds["cluster_labels"] == cluster_label)
-        .sum(skipna=True)
-        .values.item()
-    )
+    return ds["area"].where(ds["cluster_labels"] == cluster_label).sum(skipna=True).values.item()
```
BSchilperoort, line 60: Something went wrong here, Yang. These lines are >100 long. See my suggestion for the pyproject.toml file

setup.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
```
BSchilperoort, line 1: setup.py can be deleted, we don't need it anymore as we don't use setuptools.

pyproject.toml
geek-yang: Well spot! It is removed now!

s2spy/rgdr/rgdr.py
geek-yang: Solved!

setup.py
geek-yang: Good point! `setup.py` and `setup.cfg` are removed now.

pyproject.toml
```diff
@@ -1,10 +1,181 @@
 [build-system]
-requires = ["setuptools", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.publish.index]
```
BSchilperoort, line 5: Perhaps move this to after [project] to keep the [tool.hatch....] settings in one group.

pyproject.toml
```diff
@@ -1,10 +1,181 @@
...
+[tool.hatch.build.targets.wheel]
+packages = ["s2spy"]
+
+[tool.hatch.envs.default]
+features = ["dev"]
+
+[tool.hatch.envs.default.scripts]
+lint = [
+  "ruff check s2spy",
+  "mypy s2spy",
```
BSchilperoort, line 94: Why not lint and type check the tests as well?

```suggestion
  "ruff check .",
  "mypy .",
```

And then you can ignore the docstring errors for the test folder like this (see lilio for reference):
```toml
[tool.ruff.per-file-ignores]
"tests/**" = ["D"]
```

pyproject.toml
```diff
@@ -1,10 +1,181 @@
...
+  "PLR2004",  # magic value used in comparsion (i.e. `if ndays == 28: month_is_feb`).
+]
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+fixable = ["A", "B", "C", "D", "E", "F"]
+unfixable = []
+line-length = 88
+# Allow unused variables when underscore-prefixed.
+dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
+target-version = "py38"
+
```
BSchilperoort, line 154:
```suggestion

[tool.ruff.per-file-ignores]
"tests/**" = ["D"]

```

pyproject.toml
```diff
@@ -1,10 +1,181 @@
...
+  "D404",  # First word of the docstring should not be 'This'
+]
+ignore = [
+  "E501",  # Line length: fails on many docstrings (needs fixing).
+  "PLR2004",  # magic value used in comparsion (i.e. `if ndays == 28: month_is_feb`).
+]
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+fixable = ["A", "B", "C", "D", "E", "F"]
+unfixable = []
+line-length = 88
```
BSchilperoort, line 150:
```suggestion
line-length = 88
exclude = ["docs", "build"]
```

s2spy/rgdr/label_alignment.py
```diff
@@ -7,6 +8,7 @@
 from typing import List
 from typing import Set
 from typing import Tuple
+
```
BSchilperoort, line 11: Where did the empty lines between the imports come from? AFAIK our isort configuration does not do that.

See also the other .py files

pyproject.toml
```diff
@@ -1,10 +1,181 @@
...
+[tool.isort]
+py_version=39
+skip = [".gitignore", ".dockerignore"]
+skip_glob = ["docs/*"]
+force_single_line = true
+lines_after_imports = 2
+known_first_party = ["s2spy"]
+src_paths = ["s2spy", "tests"]
+line_length = 120
+
```
BSchilperoort, line 178:
```suggestion
no_lines_before = ["FUTURE","STDLIB","THIRDPARTY","FIRSTPARTY","LOCALFOLDER"]

```
This is the part that makes isort not add empty lines in the import list.

BSchilperoort (approved):
Hi Yang, are you enjoying hatch + ruff? 😄 

Please have a look at my suggestions, as some details were not configured correctly (IMO).

pyproject.toml
geek-yang: Oh, yes! Thanks for noticing it.

s2spy/rgdr/label_alignment.py
geek-yang: Thanks for the solution. Indeed I was confused at the beginning. But after some googling I found some codes actually have these empty lines. Thought I might be wrong before. It is nice that we can remove it by changing the setting. Looks much better now.

PR #171, opened 2023-07-12
Update test data in tutorial notebooks

geek-yang:
Replace test data in tutorial notebooks to be consistent with the `cookbook` repo. The data is not placed in this repo, but will be downloaded from `cookbook` when running the notebook.

*Note that the data for unit tests is not changed.


BSchilperoort (approved):
Hi Yang, it seems to work fine.

I do wonder if a different github repo is the right location to store it. Why not upload the data to zenodo? You can retrieve files the same way from there, but then we're sure it's persistent.

PR #172, opened 2023-07-27
Download example data from zenodo

geek-yang:
In all example notebooks, data is downloaded from Zenodo.

### DIFF
--- a/docs/notebooks/example_label_alignment_RGDR.ipynb
+++ b/docs/notebooks/example_label_alignment_RGDR.ipynb
@@ -21,6 +21,7 @@
                   "metadata": {},
                   "outputs": [],
                   "source": [
+                        "import urllib\n",
                         "import xarray as xr\n",
                         "import matplotlib.pyplot as plt\n",
                         "import lilio\n",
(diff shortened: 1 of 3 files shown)


BSchilperoort (approved):
Nice, seems to work.

I did notice that sadly it's not possible to pass a file object directly to xarray's netcdf4 engine :disappointed:. Now I understand why you downloaded them to disk first haha.

Peter9192:
> Nice, seems to work.
> 
> I did notice that sadly it's not possible to pass a file object directly to xarray's netcdf4 engine 😞. Now I understand why you downloaded them to disk first haha.

If you publish netcdf data file on the 4TU repo instead, you get a DOI but also opendap access :-)

geek-yang:
> > Nice, seems to work.
> > I did notice that sadly it's not possible to pass a file object directly to xarray's netcdf4 engine 😞. Now I understand why you downloaded them to disk first haha.
> 
> If you publish netcdf data file on the 4TU repo instead, you get a DOI but also opendap access :-)

Interesting, so there is some limit from zenodo side?? But I do find that with pandas you can actually load files (csv files) from zenodo directly. Leon did it in a notebook https://github.com/dianna-ai/dianna-exploration/blob/main/example_data/model_generation/season_prediction/generate_model.ipynb (check the 3rd cell). I think there is still some limit in xarray, not only due to zenodo. 😅

BSchilperoort:
The limit is in the netCDF4 library: it requires reading from an actual file on disk. OpenDAP is a protocol that allows you to access netCDF files stored on a server. Zenodo sadly does not have OpenDAP setup.

geek-yang:
> The limit is in the netCDF4 library: it requires reading from an actual file on disk. OpenDAP is a protocol that allows you to access netCDF files stored on a server. Zenodo sadly does not have OpenDAP setup.

Now I got it. Thanks Bart and Peter.

PR #173, opened 2023-08-01
Support monthly and weekly data for preprocess module

geek-yang:
Support monthly/weekly data in preprocess by providing a `timescale` argument to the user.

### DIFF
--- a/s2spy/preprocess.py
+++ b/s2spy/preprocess.py
@@ -61,9 +61,36 @@ def _subtract_trend(data: Union[xr.DataArray, xr.Dataset], method: str, trend: d
     raise NotImplementedError
 
 
-def _get_climatology(data: Union[xr.Dataset, xr.DataArray]):
+def _get_climatology(data: Union[xr.Dataset, xr.DataArray], timescale: str):
     """Calculate the climatology of timeseries data."""
-    return data.groupby("time.dayofyear").mean("time")
+    # check temporal resolution matching
+    if timescale == "monthly":
+        climatology = data.groupby("time.month").mean("time")
+    elif timescale == "weekly":
+        pass
+    elif timescale == "daily" or "hourly":
+        climatology = data.groupby("time.dayofyear").mean("time")
+    else:
+        raise ValueError("Given timescale is not supported.")
+
+    return climatology
+
+
+def _subtract_climatology(
+    data: Union[xr.Dataset, xr.DataArray],
+    timescale: str,
+    climatology: Union[xr.Dataset, xr.DataArray],
+):
+    if timescale == "monthly":
+        deseasonalized = data.groupby("time.month") - climatology
+    elif timescale == "weekly":
+        pass
+    elif timescale == "daily" or "hourly":
+        deseasonalized = data.groupby("time.dayofyear") - climatology
+    else:
+        raise ValueError("Given timescale is not supported.")
+
+    return deseasonalized
 
 
 def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
@@ -88,12 +115,23 @@ def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
         )
 
 
+def _check_temporal_resoltuion(timescale: str) -> str:
+    support_temporal_resolution = ["monthly", "weekly", "daily", "hourly"]
+    if timescale not in support_temporal_resolution:
+        raise ValueError(
+            "Given temporal resoltuion is not supported."
+            "Please choose from 'monthly', 'weekly', 'daily', 'hourly'."
+        )
+    return timescale
+
+
 class Preprocessor:
     """Preprocessor for s2s data."""
 
     def __init__(
         self,
         rolling_window_size: Union[int, None],
+        timescale: str,
         rolling_min_periods: int = 1,
         subtract_climatology: bool = True,
         detrend: Union[str, None] = "linear",
@@ -121,11 +159,14 @@ class Preprocessor:
                 climatology of the data. Defaults to True.
             detrend (optional): Which method to use for detrending. Currently the only method
                 supported is "linear". If you want to skip detrending, set this to None.
+            timescale: Temporal resolution of input data.
         """
         self._window_size = rolling_window_size
         self._min_periods = rolling_min_periods
         self._detrend = detrend
         self._subtract_climatology = subtract_climatology
+        if subtract_climatology:
+            self._timescale = _check_temporal_resoltuion(timescale)
 
         self._climatology: Union[xr.DataArray, xr.Dataset]
         self._trend: dict
@@ -149,15 +190,16 @@ class Preprocessor:
             data_rolling = data
 
         if self._subtract_climatology:
-            self._climatology = _get_climatology(data_rolling)
+            self._climatology = _get_climatology(data_rolling, self._timescale)
 
         if self._detrend is not None:
-            self._trend = _get_trend(
-                data_rolling.groupby("time.dayofyear") - self._climatology
-                if self._subtract_climatology
-                else data_rolling,
-                self._detrend,
-            )
+            if self._subtract_climatology:
+                deseasonalized = _subtract_climatology(
+                    data_rolling, self._timescale, self._climatology
+                )
+                self._trend = _get_trend(deseasonalized, self._detrend)
+            else:
+                self._trend = _get_trend(data_rolling, self._detrend)
 
         self._is_fit = True
 
@@ -179,7 +221,7 @@ class Preprocessor:
             )
 
         if self._subtract_climatology:
-            d = data.groupby("time.dayofyear") - self._climatology
+            d = _subtract_climatology(d, self._timescale, self._climatology)
         else:
             d = data
 


s2spy/preprocess.py
```diff
@@ -61,9 +61,36 @@ def _subtract_trend(data: Union[xr.DataArray, xr.Dataset], method: str, trend: d
     raise NotImplementedError
 
 
-def _get_climatology(data: Union[xr.Dataset, xr.DataArray]):
+def _get_climatology(data: Union[xr.Dataset, xr.DataArray], timescale: str):
```
BSchilperoort, line 64: You could also make timescale a literal: `Literal["monthly", "weekly", "daily", "hourly"]` ;)

s2spy/preprocess.py
```diff
@@ -61,9 +61,36 @@ def _subtract_trend(data: Union[xr.DataArray, xr.Dataset], method: str, trend: d
...
     """Calculate the climatology of timeseries data."""
-    return data.groupby("time.dayofyear").mean("time")
+    # check given temporal resolution matches with data
+    if timescale == "monthly":
+        climatology = data.groupby("time.month").mean("time")
+    elif timescale == "weekly":
+        climatology = data.groupby(data["time"].dt.isocalendar().week).mean("time")
+    elif timescale == "daily" or "hourly":
```
BSchilperoort, line 71:
```suggestion
    elif timescale in ["daily", "hourly"]:
```
See https://stackoverflow.com/a/20002504

geek-yang: Thanks for the suggestion. I think only for some very rare usecases, user might need `hourly` data for climatology. Also the function here is not grouped by hourly. I think it is better to just remove hourly.

geek-yang (commented):
For the unittest below, it fails only for python 3.8. It seems a version conflict between pandas and xarray causes the failure of resampling:

https://github.com/AI4S2S/s2spy/blob/5abf5692159d5f410c57ae11cedee0baef1b38dc/tests/test_preprocess.py#L83-L88

But since it only relates to the preparation of testing data, not about the code. I simply silent it for python 3.8. Just let me know if you have better solutions. Thanks @BSchilperoort !

s2spy/preprocess.py
```diff
@@ -1,11 +1,16 @@
...
+import warnings
+from typing import Literal
 from typing import Tuple
 from typing import Union
 import numpy as np
 import scipy.stats
 import xarray as xr
 
 
+TIMESCALE_TYPE = Literal["monthly", "weekly", "daily"]
```
BSchilperoort, line 11: I am not so sure about having this alias. Users will see this in the documentation, and when accessing the class/function using `help(...)`:
image

Perhaps this works better if you define `Timescale` as an alias ([see the python docs](https://docs.python.org/3/library/typing.html#type-aliases)).

```py
from typing import TypeAlias
Timescale: TypeAlias = Literal["monthly", "weekly", "daily"]
```
You can try that, and see if it shows up in the documentation or not.

s2spy/preprocess.py
```diff
@@ -88,12 +123,56 @@ def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
...
+            "Given temporal resoltuion is not supported."
+            "Please choose from 'monthly', 'weekly', 'daily'."
+        )
+    return timescale
+
+
+def _check_data_resolution_match(
+    data: Union[xr.DataArray, xr.Dataset], timescale: TIMESCALE_TYPE
+):
+    """Check if the temporal resoltuion of input is the same as given timescale."""
```
BSchilperoort, line 139:
```suggestion
    """Check if the temporal resolution of input is the same as given timescale."""
```

s2spy/preprocess.py
```diff
@@ -88,12 +123,56 @@ def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
...
+        "weekly": np.timedelta64(1, "W"),
+        "daily": np.timedelta64(1, "D"),
+    }
+    time_intervals = np.diff(data["time"].to_numpy())
+    temporal_resolution = np.median(time_intervals).astype("timedelta64[D]")
+    if timescale == "monthly":
+        temporal_resolution = temporal_resolution.astype(int)
+        if temporal_resolution <= 31 and temporal_resolution >= 28:
+            pass
+        else:
```
BSchilperoort, line 151:
```suggestion
        min_days, max_days = (28, 31)
        if min_days > temporal_resolution > max_days:
```
I think this is more clear and straightforward. I predefined min_days and max_days to make it a bit more explicit.

I am also not sure to what extent we should "babysit" the user. If they throw in a quarterly timeseries, I feel that's their own problem :sweat_smile:

s2spy/preprocess.py
```diff
@@ -88,12 +123,56 @@ def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
...
+            warnings.warn(
+                "The temporal resolution of data does not completely match "
+                "the target timescale. Please check your input data.",
+                stacklevel=1,
+            )
+
+    elif timescale in timescale_dict:
+        if timescale_dict[timescale].astype("timedelta64[D]") == temporal_resolution:
+            pass
+        else:
```
BSchilperoort, line 161:
```suggestion
        if timescale_dict[timescale].astype("timedelta64[D]") != temporal_resolution:
```

s2spy/preprocess.py
```diff
@@ -88,12 +123,56 @@ def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
         )
 
 
+def _check_temporal_resoltuion(timescale: TIMESCALE_TYPE) -> TIMESCALE_TYPE:
```
BSchilperoort, line 126:
```suggestion
def _check_temporal_resolution(timescale: TIMESCALE_TYPE) -> TIMESCALE_TYPE:
```

BSchilperoort (changes requested):
Hi Yang, I still have to test the code, but here are already some comments :)

BSchilperoort:
> But since it only relates to the preparation of testing data, not about the code. I simply silent it for python 3.8

It's not ideal, but I also wouldn't want to spend too much time on that. Perhaps we have to deprecate python 3.8 anyway: xarray isn't supporting it anymore for new releases.

s2spy/preprocess.py
geek-yang: Well spot 😂. Thanks!

s2spy/preprocess.py
geek-yang: Yep, this looks much nicer! Thanks!

s2spy/preprocess.py
geek-yang: Thanks for correcting me! I simply copy/paste the code from above and then edit it...Didn't pay attention. Looks so stupid 😅 .

s2spy/preprocess.py
geek-yang: Wait, just notice that `if min_days > temporal_resolution > max_days:` actually equals to ` min_days > temporal_resolution and temporal_resolution > max_days:`. It is not `or`. Should be the following:
```py
        min_days, max_days = (28, 31)
        if not max_days > temporal_resolution > min_days:
            warnings.warn(
                "The temporal resolution of data does not completely match "
                "the target timescale. Please check your input data.",
                stacklevel=1,
            )
```

s2spy/preprocess.py
geek-yang: Oh no....`TypeAlias` is available from python 3.10. In python <= 3.9 it should be imported from `typing_extensions`. (https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases)

BSchilperoort: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases it seems like sphinx autodoc does resolve the aliases (unless you explicitly don't want that). 

sphinx-autoapi (what we use) doesn't seem to support this.

semvijverberg: This is starting to look elegant :) 👀 

Showing the options (Literal['monthly', 'weekly', 'daily']) would be nice in the docs. If I understand correctly, our sphinx-autodoc does not resolve this (yet), and shinx-autoapi does? 

Is it an option to switch? or is there a downside?

s2spy/preprocess.py
```diff
@@ -88,12 +124,53 @@ def _check_input_data(data: Union[xr.DataArray, xr.Dataset]):
...
+        "monthly": np.timedelta64(1, "M"),
+        "weekly": np.timedelta64(1, "W"),
+        "daily": np.timedelta64(1, "D"),
+    }
+    time_intervals = np.diff(data["time"].to_numpy())
+    temporal_resolution = np.median(time_intervals).astype("timedelta64[D]")
+    if timescale == "monthly":
+        temporal_resolution = temporal_resolution.astype(int)
+        min_days, max_days = (28, 31)
+        if not max_days > temporal_resolution > min_days:
```
semvijverberg, line 151: Hi Yang, 

I pulled the branch to test it :) 

In my test, `temporal_resolution` for monthly data is 31. According to the statement, `max_days` (31) is not larger than 31 ;) 
`if not max_days > temporal_resolution > min_days:`

Should be:
`if not max_days >= temporal_resolution >= min_days:`

geek-yang: Oh! Yes, of course! Thanks for fixing this!

s2spy/preprocess.py
geek-yang: It is too much to switch from autoapi to autodoc simply for an unresolved alias 😂 , given that they are a bit different and autoapi works quite well for now.

I have added a hint in the tutorial notebook `tutorial_preprocessing.ipynb`:
"User needs to inform preprocessor the temporal resoltuion of their data, by setting the `timescale` argument to be either `daily`, `weekly` or `monthly` timescale."

And we have very clear error message to the user about the supported timescales:

https://github.com/AI4S2S/s2spy/blob/f157ab21ab431ca1cef93d4046abcd09c3f029f4/s2spy/preprocess.py#L131-L138

I think this should be sufficient.

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.8', '3.9', '3.10']
+        python-version: ['3.9', '3.10']
```
BSchilperoort, line 20: Can we add support for 3.11 (in a new PR)?

s2spy/preprocess.py
```diff
@@ -8,7 +8,12 @@
 import xarray as xr
 
 
-TIMESCALE_TYPE = Literal["monthly", "weekly", "daily"]
+try:
+    from typing import TypeAlias
+except ImportError:
+    from typing_extensions import TypeAlias
+
+Timescale: TypeAlias = Literal["monthly", "weekly", "daily"]
```
BSchilperoort, line 16: This sadly doesn't show up nicely in the api. I think that would require autodoc (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases). Autodoc then also requires typing out which functions/classes you want documented.

Perhaps just replace-all `Timescale` -> `Literal["monthly", "weekly", "daily"]`. A lot less work than changing the docs.

The autoapi lack of type alias support is discussed here: https://github.com/readthedocs/sphinx-autoapi/issues/224. One of the issues is that autoapi just reads the source code files, while autodoc actually imports modules. So I don't see this fixed anytime soon.

BSchilperoort (approved):
Just the annoying lack of type alias support in sphinx autoapi :/

Otherwise everything looks good.

s2spy/preprocess.py
geek-yang: > This sadly doesn't show up nicely in the api. I think that would require autodoc (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases). Autodoc then also requires typing out which functions/classes you want documented.
> 
> Perhaps just replace-all `Timescale` -> `Literal["monthly", "weekly", "daily"]`. A lot less work than changing the docs.
> 

I guess there is no better way. I will replace `Timescale` with things all spelling out.

> The autoapi lack of type alias support is discussed here: [readthedocs/sphinx-autoapi#224](https://github.com/readthedocs/sphinx-autoapi/issues/224). One of the issues is that autoapi just reads the source code files, while autodoc actually imports modules. So I don't see this fixed anytime soon.

😒 OK....too bad...

semvijverberg:
Thanks @geek-yang @BSchilperoort!

PR #176, opened 2023-11-07
Allow dask s2spy

semvijverberg:
- Fixed issue test_correlation failing because of floating point level differences. 
- dask = "allowed" for both preprocess and rgdr apply_ufunc

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
+    "dask",
     "lilio",
     "matplotlib",
     "netcdf4",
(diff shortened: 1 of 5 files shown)


geek-yang (commented):
Thanks for the PR! I think it is nice to add dask to the list. But could you please fix the test and make all the checks green?

Besides, also see my comment in issue #177 , it seems the current implementation works well with timeseries which include some `np.nan`. Could you please double check your code? Or simply add a minimum example for me to reproduce the error?

Thanks @semvijverberg !

ClaireDons:
We continued from this branch and made some changes so that the tests passed in #181, s2spy should now allow dask.

PR #178, opened 2024-01-09
dropna

pimmeerdink:


s2spy/preprocess.py
```diff
@@ -241,16 +243,20 @@ def fit(self, data: Union[xr.DataArray, xr.Dataset]) -> None:
         self._is_fit = True
 
     def transform(
-        self, data: Union[xr.DataArray, xr.Dataset]
+        self, data: Union[xr.DataArray, xr.Dataset], dropna=False
```
geek-yang, line 246:
```suggestion
        self, data: Union[xr.DataArray, xr.Dataset], dropna: bool = False
```

s2spy/preprocess.py
```diff
@@ -209,13 +209,16 @@ def __init__(  # noqa: PLR0913
         self._trend: dict
         self._is_fit = False
 
-    def fit(self, data: Union[xr.DataArray, xr.Dataset]) -> None:
+    def fit(self, data: Union[xr.DataArray, xr.Dataset], dropna=False) -> None:
         """Fit this Preprocessor to input data.
 
         Args:
             data: Input data for fitting.
+            dropna: If True, drop all NaN values from the data before preprocessing.
```
geek-yang, line 217:
```suggestion
            dropna: If True, drop all NaN values from the data along time axis before preprocessing.
```

Point out that this `dropna` works only for the time axis.

s2spy/preprocess.py
```diff
@@ -241,16 +243,20 @@ def fit(self, data: Union[xr.DataArray, xr.Dataset]) -> None:
...
 
     def transform(
-        self, data: Union[xr.DataArray, xr.Dataset]
+        self, data: Union[xr.DataArray, xr.Dataset], dropna=False
     ) -> Union[xr.DataArray, xr.Dataset]:
         """Apply the preprocessing steps to the input data.
 
         Args:
             data: Input data to perform preprocessing.
+            dropna: If True, drop all NaN values from the data before preprocessing.
```
geek-yang, line 252:
```suggestion
            dropna: If True, drop all NaN values from the data along time axis before preprocessing.
```

BSchilperoort:
Hi Pim, thanks for your contribution. However, we [modeled the .fit and .transform methods on sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html), where these can be used in a pipeline. I think it would be better to do steps like cleaning up the data separately, instead of modifying the fit/transform methods.

An alternative way of adding this functionality, while not adding arguments to the .fit and .tranform methods, would be to add an argument to the `__init__` method of the class.

geek-yang:
> Hi Pim, thanks for your contribution. However, we [modeled the .fit and .transform methods on sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html), where these can be used in a pipeline. I think it would be better to do steps like cleaning up the data separately, instead of modifying the fit/transform methods.
> 
> An alternative way of adding this functionality, while not adding arguments to the .fit and .tranform methods, would be to add an argument to the `__init__` method of the class.

Nearly forgot that when we designed this module, we would like to stay as close as possible to the `sklearn` pipeline recipe. Thanks for mentioning it!

PR #181, opened 2024-09-20
Allow dask

ClaireDons:
Continuing on from work from #176, making sure tests and formatting pass. Currently sonarcloud test is failing, probably due to changes on their end (issue #182), which we plan to fix in a future pull request.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -38,6 +38,12 @@ authors:
     given-names: Jannes
     affiliation: "Vrije Universiteit Amsterdam"
 
+    -
+    affiliation: "Netherlands eScience Center"
+    family-names: Donnelly
+    given-names: Claire
+    orcid: https://orcid.org/0000-0002-2546-4528
+
 date-released: 2022-09-02
 version: "0.4.0"
 repository-code: "https://github.com/AI4S2S/s2spy"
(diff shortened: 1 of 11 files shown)


CITATION.cff
```diff
@@ -38,6 +38,12 @@ authors:
     given-names: Jannes
     affiliation: "Vrije Universiteit Amsterdam"
 
+    -
```
BSchilperoort, line 41:
```suggestion
  -
```
the other dashes are less indented. not sure if it matters

pyproject.toml
```diff
@@ -35,6 +35,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
+    "dask",
```
BSchilperoort, line 38: Is it required as dependency? The code runs fine without dask right?

BSchilperoort (approved):
Looks good. Just two small comments

pyproject.toml
ClaireDons: Good point, I'll check it without and see if it still works.

ClaireDons: I added dask back in as the tests broke on github actions after removing, so perhaps it is needed.

